PR #34, opened 2024-10-13
supply_key_fix

lucifernix:
## About The Pull Request

Minor fix.
## Why It's Good For The Game

Its good.
## Changelog
:cl:
fix: Supply keys no longer have access to anarch properties. Fixed the warning of an useless if case also.
/:cl:

### DIFF
--- a/code/_onclick/observer.dm
+++ b/code/_onclick/observer.dm
@@ -1,5 +1,5 @@
 /mob/dead/observer/DblClickOn(atom/A, params)
-	if(check_click_intercept(params, A))
+	//if(check_click_intercept(params, A))
 		//return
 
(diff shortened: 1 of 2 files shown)


code/_onclick/observer.dm
```diff
@@ -1,5 +1,5 @@
 /mob/dead/observer/DblClickOn(atom/A, params)
-	if(check_click_intercept(params, A))
+	//if(check_click_intercept(params, A))
```
Agateophobia, line 2: You sure this won't break all of existence as we know it?

Agateophobia: Nevermind, github cut off the rest of the thing for me. Approved

Agateophobia (approved):

PR #49, opened 2024-10-14
Critical map fix

lucifernix:
## About The Pull Request
Critical fix for map problems that popped up in the last lockpick update. I wasn't able to get through the initialization process - because there were things on the map (lombard cost?) that were undefined. Seems out of nowhere, but now it works. Merge this instead of reverting lockpick skill by the way.

## Why It's Good For The Game
It is a critical fix.

## Changelog
:cl:
fix: fixed a critical issue
/:cl:


Agateophobia (approved):
Ballin B)

PR #54, opened 2024-10-16
Lockpicking final

lucifernix:
## About The Pull Request

This should (hopefully) be the last PR for lockpicking. Hope I do not jinx it. Eitherway, this does have an old commit that changes the map slightly, but I believe the newer commits from Lucia should override it. I have tried to do everything I could think of or find to get rid of that, and it simply is not working, so, I can do nuffin.
Also, this time physique factors into the rolls.

## Why It's Good For The Game
I've went over this, but lockpicking skill adds variety, as all skills do.

## Changelog
:cl:
add: Lockpicking skill.
/:cl:


Agateophobia:
No.

PR #55, opened 2024-10-16
Lockpicking skill and lock rework by @lucifernix

TheCarnalest:
## About The Pull Request
This PR completely redoes the lockpicking system, making it function off the Physical attribute and the new lockpicking ability. Lockpicking starts at zero dots and can be bought for 2 points per dot as per abilities in the tabletop. This also changes the descriptions of doors to reflect their lock difficulty unless otherwise specified, removes the Bone Key quirk since it's no longer needed, and makes lockpicking cars depend on the player's Physical + Lockpicking instead of the quirk + a 25% chance.

## Why It's Good For The Game
The old lockpicking system REALLY sucked. It was obviously a placeholder and lots of doors were just straight up unable to be picked open ever. This change makes it more accurate to the tabletop and/or videogame, along with giving players the ability to purchase their very first skill.

## Changelog
:cl:
add: A lockpicking ability that starts at zero dots and can be bought for cheap.
refactor: All doors on the map now use the new lock system.
/:cl:

### DIFF
--- a/code/modules/ziggers/car.dm
+++ b/code/modules/ziggers/car.dm
@@ -179,15 +179,20 @@
 			if(!repairing)
 				repairing = TRUE
 				if(do_mob(user, src, 20 SECONDS))
-					if(prob(25)) // TODO: implement new lockpicking mechanics
+					var/roll = rand(1, 20) + (user.lockpicking+user.physique) - 8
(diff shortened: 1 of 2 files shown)


lucifernix (approved):
Perfect. 10/10

PR #58, opened 2024-10-16
Only players can pass through umbral gateways.

TheCarnalest:
## About The Pull Request
Fixes a bug where Penumbral monsters were passing through portals and randomly appearing in places like the theatre.

## Why It's Good For The Game
No more chantry getting ripped apart by the white wolf! No more big-dicked monster popping into the material realm!

## Changelog
:cl:
fix: NPCs can no longer pass through umbral gateways. The white wolf and weird monster will stay in the Penumbra.
/:cl:

### DIFF
--- a/code/modules/ziggers/umbra.dm
+++ b/code/modules/ziggers/umbra.dm
@@ -142,8 +142,8 @@
 				M.marked = FALSE
 
 /obj/umbra_portal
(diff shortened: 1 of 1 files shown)


Agateophobia (approved):
Seems clear, good job Lucia.

PR #73, opened 2024-10-18
Cleaned up server log + added default case for Masquerade

TheCarnalest:
## About The Pull Request
Minor fixes so that observer movement doesn't constantly try to divide by 0, and NPCs don't print out a message whenever they spawn. Also makes the Masquerade value display as "stable" when it's above 1000 to prevent weird bugs with it going over. Will try to make sure it never goes over in the future, but this was just necessary to improve the code.

## Why It's Good For The Game
It's easier to see what's going wrong when there isn't 1000 runtimes every time a ghost moves! Also allows for Masquerade to still mostly work when overflowing.

## Changelog
:cl:
fix: Masquerade values above 1000 now display as stable.
refactor: No more log spam from observer moment and NPC spawning!
/:cl:


lucifernix (approved):
Looks fine, yeah.

PR #104, opened 2024-10-21
Enable disciplines while lying down

TheCarnalest:
## About The Pull Request
Adds documentation to variables used in /datum/discipline, and removes the check while casting to see whether the caster is currently knocked down. This lets people use Disciplines even when on the floor from something like Potence.

## Why It's Good For The Game
The meta is currently just knocking someone down and beating them up while they're incapable of resisting in any way. It's time to let go of that nonsense.

## Changelog
:cl:
tweak: You can now use Disciplines while knocked down. No more helplessly lying on the floor after getting Potence punched!
/:cl:

### DIFF
--- a/code/modules/vtmb/disciplines.dm
+++ b/code/modules/vtmb/disciplines.dm
@@ -1,20 +1,34 @@
 /datum/discipline
+	///Name of this Discipline.
 	var/name = "Vampiric Discipline"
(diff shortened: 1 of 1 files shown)


Agateophobia (approved):
Waow, one line fix + documentation? **SPEEDMERGE!!!!!!!**

PR #107, opened 2024-10-22
lasombra_voice_scrambling

lucifernix:
## About The Pull Request
Lasombra now can use phones normally. However, they come out completely scrambled on the other side. Spooky noises included.

## Why It's Good For The Game
Makes Lasombra cooler and more mysterious. Also can use phones when they couldn't before, but uselessly, so, still better?

## Changelog
:cl:
tweak: lasombra phone interactions
/:cl:


Agateophobia (commented):
I like this and code seems fine to me, I'll wait for Lucia to approve this though.

Plaguewalker:
Was it locally tested first, is the major thing.

PR #117, opened 2024-10-24
Fix new characters copying stats from slot 1

TheCarnalest:
## About The Pull Request
Changes character creation so that all new characters start with default stats instead of copying their stats from the first character slot.

## Why It's Good For The Game
Fixes the bug that was letting people clone their characters and basically start a new one with a jumpstart on progression, or avoid progression loss by saving their state before losing it.

## Changelog
:cl:
fix: New characters no longer have their stats copied from the first character.
/:cl:


lucifernix (approved):
Looks good. Merged.

PR #124, opened 2024-10-26
Camarilla will not get SWATted

lucifernix:
## About The Pull Request
Makes it so camarilla roles are not attacked by SWAT raids for wielding weapons.

## Why It's Good For The Game
I believe this will make playing camarilla roles a far more attractive proposition, will allow Camarilla to operate more strongly than other factions within a martial capacity. If they can walk over the streets during a Sabbat attack or a hunt for a serious masquerade breacher with their weapons - good. If they are in a factional war, this is their advantage. As is, nobody wants to play Camarilla and my intention is that playing Camarilla is both more fun and attractive.

## Changelog
:cl:
tweak: tweaked camarilla so they won't get swatted for gear
/:cl:

### DIFF
--- a/code/modules/vtmb/frenzy.dm
+++ b/code/modules/vtmb/frenzy.dm
@@ -245,7 +245,7 @@
 					var/mob/living/carbon/human/pull = H.pulling
 					if(pull.stat == 4)
 						var/obj/item/card/id/id_card = H.get_idcard(FALSE)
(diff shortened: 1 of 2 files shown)


Tonyburritos:
might be a issue if Camarilla players start really toating the line with it comes to breaking the mascarade or doing suspect stuff as i assume this doesn't just apply to the weapon breaches, but all breaches?

lucifernix:
No, it should only affect weapon's breaches, killing people on the street and whatnot still get you SWAT'ted.

PR #125, opened 2024-10-26
Disables Black Secret

Plaguewalker:
## About The Pull Request
Disables the quirk Black Secret indefinitely
test

## Why It's Good For The Game
It prevents people in the short and long-term from abusing it unless they have an actual good story reason for being cloaked with it being able to have it returned to em by admins.

## Changelog
:cl:
Black Secret has been commented out until rework occurs.


Agateophobia (approved):
We should've done that ourselves, thank you bestie.

PR #132, opened 2024-10-27
Salubri and Baali -NV Removal

Plaguewalker:
## About The Pull Request
This removes their NV capabilities which they're not supposed to have, it does not affect Baali's abilities.

## Why It's Good For The Game

Corrects and overlooked mistake that most haven't fixed yet.

## Changelog
:cl:
del: baali and salubri's nv codes


Agateophobia (approved):
Nothing it can possibly break, neither of these should have NV. Worst case scenario we add it back in.

Tonyburritos:
i think the issue with removing things from baali atm, is that its already not that appealing besides being able to shoot reskinned fireballs and having fire immunity to a extent. it doesn't feel like demon worshipers and they are not that powerful in any case. obviously its already been merged but i think without implementing fixes to the base problem of diamonion being crappy, it sorta fucks over those who spent time to make a higher generation baali.

PR #146, opened 2024-10-29
Clean up loose variables

TheCarnalest:
## About The Pull Request
Cleans a few slurs, and moves most type properties to their type definitions. This means that variables of /mob/living/carbon/human will now all be visible from its definition, instead of being strewn around the code.

## Why It's Good For The Game
It makes it easier to program or remove stuff when you actually know what's being defined for a type.

## Changelog
:cl:
code: Properties of types have been moved to be inside of their type definition instead of strewn around the code.
/:cl:


Agateophobia (changes requested):
Looks a-okay from my initial review. Can you remove the blink dagger and egorium scythe too since you already removed BKB?

PR #162, opened 2024-11-01
Disables embedding images in flavor text

lucifernix:
## About The Pull Request
Removes the ability to put an image in your flavor text.

## Why It's Good For The Game
A lot of flavor text images are just pure AI slop, or extremely large, intruding on people's examines. This can be exploited too, by malicious parties who can put anything they want there. It is also often pure cringe. Without it we would be better off.

## Changelog
:cl:
del: Removed the ability to add images to flavor text
/:cl:

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -2336,6 +2336,11 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 				if("flavor_text")
 					var/new_flavor = input(user, "Choose your character's flavor text:", "Character Preference")  as text|null
 					if(new_flavor)
+						var/pattern = "<img src="
+						var/pos = findtext(new_flavor, pattern)
+						if(pos)
+							to_chat(src, "Embedding images is not allowed.")
+							return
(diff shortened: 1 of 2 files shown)


Odisurin:
This can be easily moderated though?.. Like it helps with knowing what characters look like.

CMDR-Gungnir:
I'm against this as well. We can enforce any rules we want to have about it, and set a maximum size that people have to abide by, instead.

MercuryArrow:
I have to agree with the others - this is a fairly well-liked feature and there's no particular need to employ a code-based solution to what is otherwise a simple moderation problem, especially when people can put in pretty big paragraphs of text and eat space that way as well.

Peppercorn-Z:
I am also against this change for the most part, I think that people have been spreading the information of how to do this without additional context like height= and width=. AI art being slop is a rather subjective point I have spent a substantial amount of money over the years for commissions of the characters I play, and currently have a commission in progress for my Baali. Having ai art to stand-in while I save up to afford portraits is nice.

wod13-server:
#### Test Merge Removed

PR #171, opened 2024-11-03
Adds fat versions of the Primogen Toreador clothes.

punchy6700:
does what it says on the tin


lucifernix (approved):
No reason to reject this...

PR #174, opened 2024-11-04
disables DoC ambient loop

YuiY1997:
## About The Pull Request

changes the TRUE variable in ambient.dm to stop the cacophony ambient loop from playing

## Why It's Good For The Game

This curse provides no IC drawbacks and is purely an OOC annoyance. It also did not function like other music in the game and could not be disabled by any means like ambient music or lobby music.

## Changelog
:cl:
tweak: Changed the variable in Ambient.dm to disable the daughters of cacophony sound loop

/:cl:

### DIFF
--- a/code/modules/wod13/ambient.dm
+++ b/code/modules/wod13/ambient.dm
@@ -427,7 +427,7 @@
 				var/mob/living/carbon/human/H = src
 				if(H.clane)
 					if(H.clane.name == "Daughters of Cacaphony")
-						cacaphony = TRUE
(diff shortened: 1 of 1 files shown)


YuiY1997:
if this doesn't work throw tomatoes at me in the stocks

MercuryArrow:
The funny thing is that the DoC curse is actually a benefit as it overrides most other music loops, meaning DoC don't get the brief moment of lag when switching areas and loading in a new song.

PR #199, opened 2024-11-12
LATE PARTY 2 

Plaguewalker:
## About The Pull Request

Reduces the cycle time dramatically, and condenses the late roles down into the base 5 removing the two extra dog spawns.

## Why It's Good For The Game

Overly padded with BSD spawn get cut down allowing for more late parties to cycle

## Changelog
:cl:

tweak: tweaked a few things

/:cl:


code/modules/vtmb/gamemodes/subsystem.dm
```diff
@@ -5,7 +5,7 @@
 SUBSYSTEM_DEF(bad_guys_party)
 	name = "Bad Guys Party"
 	init_order = INIT_ORDER_DEFAULT
-	wait = 12000
+	wait = 2000
```
lucifernix, line 8: 12000 to 2000 is a little drastic, change it to 400 and we can go with that IMO.

Plaguewalker: > 12000 to 2000 is a little drastic, change it to 4000 and we can go with that IMO.

Changed it

lucifernix (approved):
I think this is a fair change that will make rounds more interesting... Approved.

PR #215, opened 2024-11-16
Furious Garou

CatandWet:
## About The Pull Request
THE BUFF THAT SAVED AN ENTIRE SPECIES
Makes garou not be affected by the miss chances in melee that continue to stack up as you get more wounded, making them rightfully more capable in their niche without making them too much tougher/sturdier/more ridiculous than they already are.

## Why It's Good For The Game
A wolf forced into a corner should bite viciously not whiff 75% of the hits that matter.

## Changelog
:cl:
add: One little trait
/:cl:


lucifernix (approved):
Seems solid.

PR #228, opened 2024-11-22
{OLD STALE}Wolf Hardiness (to very niche insta-kills) [DON'T MERGE]

CatandWet:
## About The Pull Request
Makes a garou's experience a little less mindlessly abuseable to the point of instantly keeling over when a vicissitude user walks on screen; now the wolf must be fought instead of ceasing to function from various odd occurences. Garou blood regen still may need a look at so thauma/serpentis/other blood stealing discs can work on wolves better.

+ Lessens MURDER impact of vicissitude dramatically, now does brute and clone to a garou.
+ Thaum steals blood + bolts do burn as temp solution but shouldn't outright kill (not sure it ever did)
+ Gives garou rage heal oxy damage healing for other survivability edge cases

## Why It's Good For The Game
More wolf fixes.

## Changelog
:cl:
add: more teeth
del: bad howling habits
tweak: muscle density
balance: fang sharpness
fix: garou being pushovers
/:cl:


CatandWet:
Tested a good dozen times, vomitting up blood/stealing/other bloodpoint related attacks can send a garou into negative bloodpoint values but does not seem to kill or bother them? Unsure of how to make it so if a garou gets emptied that hard their bloodpoints reset to zero and start regening but otherwise this all seems to work.

code/modules/vtmb/disciplines.dm
```diff
@@ -1200,18 +1203,23 @@
 
 /datum/discipline/vicissitude/activate(mob/living/target, mob/living/carbon/human/caster)
 	. = ..()
-	if(iswerewolf(target))
+	if(iswerewolf(target) || isgarou(target))
 		caster.playsound_local(caster.loc, 'code/modules/wod13/sounds/vicissitude.ogg', 50, TRUE)
-		caster.adjustFireLoss(50)		//abusers suffer
+		caster.adjustFireLoss(35)		//abusers suffer
+		caster.Stun(30)
 		caster.emote("scream")
```
Agateophobia, line 1210: Is punishing the caster necessary if this works fine now? 🤔

code/modules/vtmb/disciplines.dm
```diff
@@ -1292,8 +1300,8 @@
 				var/obj/item/bodypart/B = H.get_bodypart(pick(BODY_ZONE_R_ARM, BODY_ZONE_L_ARM, BODY_ZONE_R_LEG, BODY_ZONE_L_LEG))
 				if(B)
 					B.drop_limb()
-	else
-		target.death()
+	//else
+		//target.death() - Removed until a better solution is found to not have insta-kills on player mobs, unsure of side effects for normal vicissitude use but call death above already so should be fine?
```
Agateophobia, line 1304: This is basically "If it's not human, not kindred and not garou, just kill the thing", so it shouldn't cause any problems, no?

CatandWet: That's what I thought.. and yet it seemingly WAS the wolf killer; I tried a few ways to keep it in and exempt wolves species from the rest of it but couldn't get it to compile or work right, whereas just removing this.. seemingly had no repercussions since npcs, humans, kindred and garou were all accounted for.. at least that was my understanding of how it all worked, I could be completely wrong.

code/modules/vtmb/disciplines.dm
CatandWet: The little bit of feedback to the caster to prevent spamming felt good in testing but I'll turn down the damage, it is still 70 fire dam for a kindred which.. is absurd when it's no longer needing punishment.

Agateophobia: I would remove it entirely tbh. It's super unnecessary, you could probably also remove stuns from both sides.

Agateophobia:
You want me to like, close this?

PR #233, opened 2024-11-23
Gridventory bag limiting

CatandWet:
## About The Pull Request
Keeps satchels at the 6x6 standard while improving backpacks and duffels back to similar standard they were on old weight class system.
Find out in game if you think your favourite item is too large or small to fit in a satchel and yell at me until it's changed.

## Why It's Good For The Game
No more carrying AR's in a satchel

## Changelog
:cl:
tweak: tweaked inv values I dunno look at the changes
/:cl:


cornercube:
you made the uzi mag 1x1, it being 1x2 was probably deliberate to reign in automatic fire so it isnt strictly better than the glock

PR #240, opened 2024-11-24
Serpentis / Quietus Adjustments

Peppercorn-Z:
## About The Pull Request
This makes the serpentis urn able to be picked up by people, previously simply touching it would kill the creator.

This also makes it so that quietus 1 doesn't create an overlay around every single person it applies to.

## Why It's Good For The Game
Being able to handle the urn creates a ton of roleplay opportunity, such as holding a ministry character's urn as insurance of a life boon, and actually enables the serpentis users to properly hide the item in containers.

Quietus 1 not creating an overlay on everyone it's active on makes it a more stealthy discipline while also requiring more tactful use from the banu using it.

## Changelog
:cl:
tweak: tweaked the way quietus 1 operates, in regards to mutation overlays.
fix: fixed serpentis urn not being able to be handled by players, it can now be destroyed by using in-hand or attacking with another object.
/:cl:


code/modules/vtmb/vampire_clane/ministry.dm
```diff
@@ -30,35 +30,34 @@
...
 
-/obj/urn
+/obj/item/urn
 	name = "organ urn"
 	desc = "Stores some precious organs..."
 	icon = 'code/modules/wod13/icons.dmi'
 	icon_state = "urn"
-	plane = GAME_PLANE
-	layer = CAR_LAYER
 	var/mob/living/own
```
Agateophobia, line 38: A little late review, but does the urn display properly when dropped, etc? Otherwise this seems good to go.

Peppercorn-Z: Yep yep. The plane and layer are for more structure type items, which the urn is not anymore.

PR #243, opened 2024-11-27
Places scourge under the master file

AllCorruptingWyrm:
## About The Pull Request

Scourge is a generally unrewarding job, and I thought it was really weird that the role doesn't get it given how important it is to the health of the masquerade and, by extension, the round.

## Why It's Good For The Game

This PR just rewards Scourges for playing their role and actually want to maintain it, instead of just jumping to Sheriff.

## Changelog
:cl:
fix: Scourges now gain additional XP
/:cl:


Agateophobia (commented):
You should adjust the changelog, as it's a player-facing thing. Something like "Scourges now gain additional XP" would be fine.

PR #250, opened 2024-11-28
Temporal Tampering - Bullettime

Peppercorn-Z:
## About The Pull Request
Adds bullet dodge chance to temporis 4, and full bullet dodge to temporis 5

## Why It's Good For The Game
Right now temporis 4 is virtually a useless ability, unless someone is really awful at clicking on you. This gives it utility it did not have before, while still not being perfect. Temporis 5 is not clearly advertised enough as being separate from celerity, and is not a risk to use in public currently, now, if a true brujah uses it in public and is shot at, they will 'step out of the way' of bullets, revealing themselves to some degree. Temporis 5 is also the most expensive power in the game currently, bloodpoint wise.

## Changelog
:cl:
add: Added bullet dodge abilities to some temporis levels
/:cl:

### DIFF
--- a/code/datums/martial/cowalker.dm
+++ b/code/datums/martial/cowalker.dm
@@ -1,0 +1,19 @@
+/datum/martial_art/cowalker
+	name = "Cowalker"
+	id = MARTIALART_COWALKER
+	allow_temp_override = FALSE
+
+/datum/martial_art/cowalker/on_projectile_hit(mob/living/A, obj/projectile/P, def_zone)
+	. = ..()
+	if(iskindred(A) && A.temporis_visual == TRUE)
+		if(prob(75))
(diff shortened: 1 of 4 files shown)


code/modules/vtmb/vampire_clane/true_brujah.dm
Peppercorn-Z, line 1: reduced clotho's cooldown time to allow trujah the time old ability to temporis themselves into a husk, and bring it more into line with some of the other super stronk abilities from other clans (like shadow/blood armor)

code/datums/martial/cowalker.dm
Peppercorn-Z, line 1: Yes, this was the easiest way to add these to the disciplines, as silly as that is. I didn't want to make two separate martial arts for the two different disciplines it affects, and simply named it after the first affected, in order of acquisition

PR #259, opened 2024-11-30
Experiencening 2.0

Peppercorn-Z:
## About The Pull Request
Effectively adds the experience bonuses that Lucifernix wanted added

## Why It's Good For The Game
Cuts down on some of the XP grind, and makes less painful some of the XP expenditures planned for the future

## Changelog
:cl:
tweak: tweaked the way experience is calculated per job, and the time of night that experience is calculated.
/:cl:


code/controllers/subsystem/city_time.dm
```diff
@@ -53,8 +53,14 @@ SUBSYSTEM_DEF(city_time)
...
+									P.add_experience(8)
+								if("[H.mind.assigned_role]" == "Seneschal" || "[H.mind.assigned_role]" == "Bogatyr" || "[H.mind.assigned_role]" == "La Squadra" || "[H.mind.assigned_role]" == "Graveyard Keeper" || "[H.mind.assigned_role]" == "Doctor" || "[H.mind.assigned_role]" == "Janitor" || "[H.mind.assigned_role]" == "Archivist" || "[H.mind.assigned_role]" == "Supply Technician")
+									P.add_experience(7)
+								if("[H.mind.assigned_role]" == "Scourge" || "[H.mind.assigned_role]" == "Triad Soldier" || "[H.mind.assigned_role]" == "La Famiglia" || "[H.mind.assigned_role]" == "Taxi Driver" || "[H.mind.assigned_role]" == "Bouncer" || "[H.mind.assigned_role]" == "Stripper")
+									P.add_experience(4)
```
Arturlang, line 63: At this point you probably should just check a list of roles instead of a huge if statement since you have so many, easiest way would be to have
var/list/mylist = list("Taxi Driver", "Bouncer")
if(H.mind.assigned_role in mylist)

Arturlang: though, even better would be just using a job typepath, but then you'd have to do check it in a foreach, so maybe avoid that

faaaay: Would it be viable to repurpose the experience_added variable on each job? Say, each job has a var defining how much XP they get per interval, which can then be numerically modified in the proc. (I.e. 1.5 times as much at roundend)

Peppercorn-Z: That's fair. I have mostly gone hands off with this PR unfortunately, as the main dev of the project has vastly different XP values in mind (much lower, but I do not know the exact numbers), but they will be able to see these comments 👍

Agateophobia:
I'll just close this, because it's:
a) stale
b) similar PR was closed as well

PR #273, opened 2024-12-02
CHUNK TIME

cornercube:
## About The Pull Request
adds a single spawner to millenium apartments that allows an observer to play as the millenium towers security guard

basically chunk RP, his keys open every room in the building except for prince-only rooms
he can man the front desk to screen people for the prince
he is minimally armed, a single M1911 and 5 donuts

a fluff role, he is effectively useless and he will probably end up dead before the night is over
but it is funny
his keys have more access than the camarilla scourges, he could potentially allow them audience with the prince

also touches up the generic guard to use the M1911 as well and to have more unique dialogue
some of which were inspired by chunk

## Why It's Good For The Game
image

be the only thing holding back chaos from getting to your boss

## Changelog
:cl:
add: chunk role
/:cl:


cornercube:
it should work now

image

PR #323, opened 2024-12-13
Gives more clans the keys to their havens

faaaay:
## About The Pull Request

Gives Brujah, Malkavian, Nosferatu and Toreador the keys to their clan havens upon spawning so that they don't have to break in. Slightly changes how this is handled so that instead of it being added to the `post_gain` proc, any clan can have their `clan_key` variable set to make sure they spawn with a certain item.

## Why It's Good For The Game

Not everyone builds for lockpicking, and these areas go underused in part because it's inconvenient to get to them. Nosferatu specifically benefit strongly from having access to the hideout they spawn next to, because of the mask vendor inside. A primogen can potentially let you in, but that's inconsistent at best and (for Nosferatu) rarely seen at worst.

## Changelog
:cl:
tweak: brujah, nosferatu, malkavians, toreador all spawn with the keys to their clan hideout
code: changed how clan keys are spawned behind the scenes
/:cl:

### DIFF
--- a/code/modules/vtmb/vampire_clane/baali.dm
+++ b/code/modules/vtmb/vampire_clane/baali.dm
@@ -11,13 +11,13 @@
 	female_clothes = "/obj/item/clothing/under/vampire/baali/female"
 	enlightenment = TRUE
 	whitelisted = TRUE
+	clan_keys = /obj/item/vamp/keys/baali
 
 /datum/vampireclane/baali/on_gain(mob/living/carbon/human/H)
 	..()
 	H.faction |= "Baali"
 	var/datum/brain_trauma/mild/phobia/security/T = new()
 	H.gain_trauma(T, TRAUMA_RESILIENCE_ABSOLUTE)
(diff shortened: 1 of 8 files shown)


faaaay:
Haven't tested this with every clan, but I spawned as Nosferatu Citizen on a local server and it worked fine.

wgtjunior743:
Nosferatu primogen area is really good for that change, but other primogens areas just got nothing besides some stuff for the primogen and are too small for any realgathering, for the good side if there are not primogens you can steal their stuff more easily for the bad side, the primogen might get mad with people easily getting into their places especially if a blood hunt,ed sabbat or even someone that dislikes them get easy acess to their hiding places.

faaaay:
The Primogen offices are still locked for only the Primogen. There's not a whole lot for some clans to do with their haven, but the haven is there for them; it'll just go unused if they don't have access to it. If someone wants to break in, they can already do so by force.

PR #331, opened 2024-12-14
Update to Phones, add Remove, Block , Unblock, Important contacts also now got their names.

wgtjunior743:
## About The Pull Request

Finnaly now you can block that annoying caller or remove a contact that you may have wrote wrong a number or name.
Also as a Prince or others you don't need to guess peoples name now, they will have their names right attached to the contacts once they open their phone ( and the important contacts also opened theirs)
I was going to make for a unpublish but it makes more sense if you fuck up that.
Silence option and caller historic will come depending on how fast this is merged.

Compilled tested and tested ingame if it works.

Fixed the conflict merges.

## Why It's Good For The Game

More options to the phone, it is not near perfection but got more stuff that it should have since the start.

## Changelog
:cl:
add: Important Contacts got names attached to it now.
add: Remove Contacts option.
add: Block and Unblock options.
/:cl:

### DIFF
--- a/code/_globalvars/phonesnumbers.dm
+++ b/code/_globalvars/phonesnumbers.dm
@@ -1,3 +1,5 @@
+// Important Numbers
+
 GLOBAL_VAR_INIT(princenumber, "")
 GLOBAL_VAR_INIT(sheriffnumber, "")
 GLOBAL_VAR_INIT(clerknumber, "")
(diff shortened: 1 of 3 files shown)


lucifernix (approved):
This seems good to me. I tested it and it seems to be working.

PR #337, opened 2024-12-15
permanent chunk role

cornercube:
## About The Pull Request
makes chunk a real job
neutral-aligned, human-only role
still a joke role: so other than access to most of the tower, he has no real gear or special features
mostly the same as the spawner role, it just allows you to have a consistent name

he also gains a red desk phone that can call a matching phone in the camarilla office
so he has further ability to annoy the prince with guests
intruders who kill chunk can use the phone to send ominous threats to the prince

## Why It's Good For The Game
a role with no real impact, but gives human mains another thing to do
which is annoy the camarilla

makes red landlines have an old fashioned ringing bell noise

## Changelog
:cl:
add: chunk role (millenium tower security) 
add: camarilla penthouse and reception phones
add: oldfashioned cameras ring sound for landlines
fix: police chief spawn
/:cl:


cheapcents:
Hey there,
we already have a millennium tower security guard ghost spawn so i think this PR is redundant. It'd be better if we imported the ghost spawners menu from tgstation so that more people would be aware of it, and then modified the ghost spawn itself to reflect this PR, if it isn't good enough

cornercube:
> we already have a millennium tower security guard ghost spawn so i think this PR is redundant.

I made the ghost spawn role as a test for whether people would like it as a permanent role, and people also want to be able to have a static name when playing as it

PR #359, opened 2024-12-18
Scrap Independent, new Choose profession menu and new colors for occupation preferences

wgtjunior743:
## About The Pull Request

Choose profession now actually looks good and it is organized, i always disliked like everything was threw inside independent and became a big blob.

New Choose profession menu:

new_choose_profession

New Occupation Preferences menu:
new_occupation_preferences_2

I know, this is going to conflict soon with some prs when they get merged, i will deal with it.

## Why It's Good For The Game

Better looking, easy to pick and see to which groups your profession is part of.

## Changelog
:cl:
del: some of ss13 stuff
refactor: old ss13 stuff blob became a single glob called ss13
add: new globs for each group/factions/family
/:cl:


AllCorruptingWyrm:
HOLY FUCK I LOVE YOU

does this mean we can start adding jobs here and the job code categories are all modularized? I'm trying to appraise the code.

wgtjunior743:
> HOLY FUCK I LOVE YOU
> 
> does this mean we can start adding jobs here and the job code categories are all modularized? I'm trying to appraise the code.

For those categories created, yes you can add jobs under them and use the same exp_type, but the display has to be unique, but if the job don't fits any of the categories already created you can make your own, preferences.dm > #define EXP_TYPE_(name) "(name)", then jobs.dm GLOBAL_LIST_INIT((name)_positions, list( ("new job"),))  , then just add position_categories for the color and order and then to exp_jobsmap so they can appear right.

PR #367, opened 2024-12-21
Adds old clan tzimisce up to be whitelisted

AllCorruptingWyrm:
## About The Pull Request

This seems to be an oversight but one that really should be fixed. In numerous incidents I've witnessed, I've seen old clan tzimisce be complete and utter shitters with their OP armor with the most lore breaking names imaginable. This fixes it.

## Why It's Good For The Game

One of the main things is that Old Clan are kind of rare, and they get stupidly good armor which they use on a consistent basis to morb out.

## Changelog
:cl:
tweaks: Makes old clan tzimisce whitelisted
/:cl:


lucifernix:
There are currently no plans to whitelist Old Clan Tzimisce as the most you need to know to RP them properly is to roleplay as Dracula.

PR #372, opened 2024-12-22
Normalises XP rates a little

faaaay:
## About The Pull Request

The basic rate of XP per-XP-check (when the game checks for social, cleaning and base XP) is now 2, not 1, and leadership roles (i.e. baron, sheriff, dealer) now get +2 XP rather than +3. Social XP is up by 1 across the board to 2 per check, and the Toreador bonus has been lowered to 1 from 2. 
In other words: Toreador kindred and leadership roles will get the same amount of XP, but anyone else will be a little closer to the frankly-disproportionate XP rates (a Toreador Seneschal getting 21 XP when someone doing the same amount of RP would get **6**).

## Why It's Good For The Game

"Won't this hasten ~~Gehenna~~ the elderpocalypse by giving out more XP and letting people become super-powerful now?"
Not really. Someone who really wants to grind XP can do that with relative ease already, and this doesn't make it any easier for them. But equalising other rates _up_ avoids encouraging people to jump into roles they're not ready for and play a plan they're not interested in to progress at a rate that feels meaningful.

## Changelog
:cl:
tweak: base xp per round is now 6, +6 over the whole round for leadership roles. this is a baseline increase and makes no difference to roles that get extra xp
tweak: social xp is now 2 per shot, with the toreador bonus decreased by 1. this is a baseline increase and makes no difference to toreador characters.
/:cl:

### DIFF
--- a/code/controllers/master.dm
+++ b/code/controllers/master.dm
@@ -250,10 +250,10 @@ GLOBAL_REAL(Master, /datum/controller/master) = new
 					if(H.key)
 						var/datum/preferences/P = GLOB.preferences_datums[ckey(H.key)]
 						if(P)
-							P.add_experience(1)
+							P.add_experience(2)
 							if(H.mind)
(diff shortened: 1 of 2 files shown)


faaaay:
Updated at Weaver's request:

- Base XP back at 1 (per check - assume all values are per check, not per round))
- Leadership XP down to 2
- Social XP set at 2 for everyone, including Toreador. No more clan XP bonus.

PR #378, opened 2024-12-23
Fix Malkavians not getting their keys.

wgtjunior743:
## About The Pull Request

[Compiled tested and tested ingame]

Previously malkavian did not spawned with their clan keys unlike the others clans with a primogen, it was not really an issue with the code just that who made the malkavian post gain proc forgot to put the "..()" so it was not inheriting the clan.dm proc and just overwriting with their own.

Le keys:
image

## Why It's Good For The Game

Fix!

## Changelog
:cl:
fix: Fix malkavians not getting their hideout keys.
/:cl:


theXDkidoflol:
speedmerge immediately

me and my siblings in madness need a place to take all our bitches in madness

PR #382, opened 2024-12-24
Fix Ghoul aura oversights

wgtjunior743:
## About The Pull Request

i aimed to fix two oversights that i don't checked on the Ghoul aura PR, now if you are a human that is transformed into a ghoul your aura will now be ghoul aura(holder.color and overlays were making it still be blue)
If a vampire transform an npc in a ghoul and a player accept as well, even if they are harm intent their aura won't be black-red anymore, it will become a normal ghoul aura.

## Changelog
:cl:
fix: Turned ghouls mid round aura is now ghoul aura
fix: Turning npcs into ghouls will always give them ghoul aura (they only become a ghoul if a ghost accepts)
/:cl:

### DIFF
--- a/code/modules/wod13/witness.dm
+++ b/code/modules/wod13/witness.dm
@@ -62,15 +62,22 @@
 		var/ghoul = FALSE
 		if(isghoul(src))
 			//Pale spots in the aura, had to be done manually since holder.color will show only a type of color
+			holder.overlays = null 
+			holder.color = null
 			holder.icon_state = "aura_ghoul"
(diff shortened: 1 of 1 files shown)


wgtjunior743:
> This code looks... strange. I'll have to try to correct it later, but it's very difficult to read.

that is true, i updated with a better logic like i had implemented for players

wgtjunior743:
> The code needs some neatening. I added a comment for that, just look at it and do your thing then we can merge!

very nice, thanks, i tested and it is good too.

PR #390, opened 2024-12-26
Window climbing improvements

boy2mantwicethefam:
This took way longer than it should have.

- Fixed a bug where it was impossible to climb through low wall windows because a wall's loc is the area it is in and not itself.
- Step() was awkward, and it was replaced with Move(), which also allows for climbing low walls diagonally.
- If you climbed on a low wall then you can move to adjacent low walls and tables, and vice versa.
- Prevented players from climbing something they're already on.
- Fixed a bug where low walls blocked air movement even when they shouldn't. This meant that a character could suffocate while on a low wall.

No changelog because the changelog system doesn't work yet.

### DIFF
--- a/code/datums/elements/climbable.dm
+++ b/code/datums/elements/climbable.dm
@@ -34,6 +34,9 @@
 		examine_texts += "<span class='notice'>[source] looks climbable</span>"
 
 /datum/element/climbable/proc/can_climb(atom/source, mob/user)
+	if(get_turf(user) == get_turf(source))
+		to_chat(user, "<span class='warning'>You are already on \the [source]!</span>")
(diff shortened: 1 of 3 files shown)


lucifernix:
This seems to create some issues with... the method of climbing on top of walls. In particular, you get things like this 
image

Other than that I didn't find any issues, but it's awkward that it starts both of the climbing processes at once.

boy2mantwicethefam:
That issue has been there since wall climbing has been a thing, not sure how to tackle that other than require wall-climbing to be on grab intent, but that'd need to be noted down somewhere in-game to make it more user-friendly

PR #394, opened 2024-12-27
Tiny changes

boy2mantwicethefam:
### Main features
- Archetypes will now feature text that says what their actual bonus ability is.
- Bonus attributes will now be visible on the character screen and when checking your own stats.
image
image
- Fixed a bug where the Dude archetype did not actually work in affecting ranged weapon critical chances. It will now properly provide +3 Cruelty to critical damage calculations.
- Fixed a bug where the Cruelty attribute's upgrade cost text didn't reflect its proper cost, which was supposed to be doubled.

### Back-end stuff

- Reworked some backend stuff to make a few bits of the attribute system more modular and reduce copypaste. The system will now support displaying more than 5 dots on the character menu.
- Added support for bonus attributes; dexterity, lockpicking and athletics can now also benefit from the system, but currently nothing makes use of it. Code elements that only made use of those base attributes will now also consider bonus attributes.
- Added support for archetypes being able to contribute to all 7 attributes rather than just 4 (Physique, Social, Mentality, Cruelty).
- Added helper procs for getting the total amount of stats (attribute + bonus attribute), which should help tidy things up a bit.

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -471,7 +471,7 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 			var/dexterity_price = 4
 			var/social_price = 4
 			var/mentality_price = 4
-			var/blood_price = 3
+			var/blood_price = 6
 			//Lockpicking and Athletics have an initial price of 3
 			var/lockpicking_price = !lockpicking ? 3 : 2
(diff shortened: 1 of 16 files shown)


Agateophobia (approved):
Great PR. Approved and Merged.

PR #396, opened 2024-12-27
Frenzy tweaks + fearless to fire kindred fix (unfunctional)

CatandWet:
## About The Pull Request
Poorly and shoddily attempts to patch up frenzy code and rotshrek's old step away when being too close to a fire (which when it worked was too severe and annoying)
Most changes regarding fearstack and the uncommented back in blood frenzy tested and seem to be working (?) and values already lowered so it's not a constant thing. 

It's mostly the fear atom and/or step away being the issue I can't solve, so please delete my awful attempt to fix it and do it better or yell at how it should be done.

+ Also lowers saulocept XP gains since the xp hungry should not be so well fed.

## Why It's Good For The Game
Kindred can stand right next to huge infernos without flinching right now.

## Changelog
:cl:
tweak: tweaked a few things
balance: the range NPCs spot frenzy-iers
fix: barely anything I need assistance

/:cl:

### DIFF
--- a/code/modules/vtmb/artifacts.dm
+++ b/code/modules/vtmb/artifacts.dm
@@ -101,7 +101,7 @@
 
 /obj/item/vtm_artifact/saulocept/get_powers()
 	..()
-	owner.experience_plus = 10
+	owner.experience_plus = 5
 
 /obj/item/vtm_artifact/saulocept/remove_powers()
 	..()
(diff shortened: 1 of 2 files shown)


Metekillot:
I support what you're doing here but this code needs a full rewrite and not a patch job. No more stopgaps, because stopgaps on top of stopgaps leads one to Wassail

PR #398, opened 2024-12-27
Add a choice to keep being a vampire after a embrace or after being turned into a ghoul

wgtjunior743:
## About The Pull Request

After being transformed into a vampire/ghoul, now you can pick between "Yes" or "No" to save that change on your slot, keep in mind that WL clans for those without the WL, will be changed to their non WL versions or randomized.
For power gaming reason, if your sire is below the 13 gen, you will always be a 13th gen as well, if they are 13th gen you will be turned into a Caitiff 14th gen but unfortunately the game by default will force the 14th gen to turn 13th on the char slot.

On a side note for coders, create_disciplines does not work if you pass a list, i tried a couple of time to fix but it was not my focus so i used give_discipline directly, that also means that gargoyles created ingame don't recieve their disciplines.

Vamp choice:
vamp_ask

Ghoul choice:
ghoul_choice

## Why It's Good For The Game

This is mainly for rp reasons, so people call roleplay as real sire/child or turn a human into a ghoul and they choose to keep,
also WLs can make their story based of a real player as their sire and that can happen mechanically.
Also someone who plays human alot and spent alot of xp on it can turn into a vampire/ghoul without the need to reset their stats.

## Changelog
:cl:
Add: A choice to keep the vampire change on your slot after being transformed
Add: A choice to keep the ghoul change on your slot after being transformed.
Fix: The child of a embrace now gain their skills
/:cl:

### DIFF
--- a/code/modules/vtmb/kindred_species.dm
+++ b/code/modules/vtmb/kindred_species.dm
@@ -399,6 +399,7 @@
 						log_game("[key_name(H)] has Embraced [key_name(BLOODBONDED)].")
 						message_admins("[ADMIN_LOOKUPFLW(H)] has Embraced [ADMIN_LOOKUPFLW(BLOODBONDED)].")
 						giving = FALSE
+						var/save_data_v = FALSE
 						if(BLOODBONDED.revive(full_heal = TRUE, admin_revive = TRUE))
 							BLOODBONDED.grab_ghost(force = TRUE)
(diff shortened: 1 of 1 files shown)


lucifernix:
I don't think it's desired that people can just embrace characters and have them be their gen + 1, but I'll leave this one up to @TheCarnalest

wgtjunior743:
> I don't think it's desired that people can just embrace characters and have them be their gen + 1, but I'll leave this one up to @TheCarnalest

their gen is always set to 13 no matter what, if the sire is gen 7 their childe will be always 13, i did that so ppl do powergame

PR #409, opened 2024-12-29
Vampires now take extra damage from being on fire.

boy2mantwicethefam:
For every tick (about 2 seconds) that a vampire is on fire they additionally take 4% of their max health in burn damage, with their weakness for fire already accounted for (the actual number is 2%). This is merely max health and NOT the total health of a vampire, which is an additional 200 while in a critical state (humans can get to -100, vampires can get to -200 but the last 100 are in torpor).

For some numbers, a theoretical 7th gen vampire with 6 Physique would have a max health of 375, which would make them take an additional 15 burn damage per tick while they're set on fire. It is effectively dealing +1 damage for every 25 max health points.


Plaguewalker:
They already take double burn damage in general.

code/modules/vtmb/kindred_species.dm
```diff
@@ -710,3 +710,9 @@
 			log_game("[key_name(teacher)] taught [key_name(student)] the Discipline [giving_discipline.name].")
 
 		qdel(giving_discipline)
+
+//Vampires take 4% of their max health in burn damage every tick they are on fire. Very potent against lower-gens.
+//Set at 0.02 because they already take twice as much burn damage.
+/datum/species/kindred/handle_fire(mob/living/carbon/human/H, no_protection)
+	if(!..())
+		H.adjustFireLoss(H.maxHealth * 0.02)
```
Metekillot, line 718:
```suggestion
		H.adjustFireLoss(H.maxHealth * 0.10)
```
Being on fire for a vampire is an apocalypse-level "holy fuck" moment. Being lit on fire as a vampire should, in fact, also make the player go holy fuck. Being lit on fire as a vampire is one of the absolute last things that you want to happen to you ever and you should be going ridiculously out of your way to avoid it.

Metekillot (changes requested):

code/modules/vtmb/kindred_species.dm
boy2mantwicethefam: They'd lose 20% max health every 2 seconds. Not a fun thing for the player behind the vampire to immediately collapse in a few seconds from one incendiary bullet.

Metekillot: > They'd lose 20% max health every 2 seconds. Not a fun thing for the player behind the vampire to immediately collapse in a few seconds from one incendiary bullet.

I'll meet you in the middle at 0.05 then. 10% every 2 seconds is BEEF, but if they think fast they MIGHT survive. Being lit on fire should be the absolute high-end oh shit factor for a vampire.

Metekillot, line 718:
```suggestion
		H.adjustFireLoss(H.maxHealth * 0.05)
```

Metekillot (changes requested):

Metekillot (approved):
after several of the admin team informed me that I was in fact Full Of Shit regarding my understanding of how much damage vamps take from fire in the lore, this is actually good to go.

Metekillot (approved):
after several of the admin team made it apparent to me that I was in fact Full Of Shit regarding what i thought i knew about vamps taking fire damage, this actually looks good

PR #414, opened 2024-12-30
Fixes atmospherics initializations

boy2mantwicethefam:
- Fixed a lot of red carpets throughout the map spawning with no air.
- Fixed some tiles in the Siren Cabaret spawning with no air.
- Tech shop server room is now at room temperature.
- Hopefully patched up all the instances where tiles would spawn with different air, which would significantly slow down atmospheric initializations as it would attempt to normalize the air between the tile and other tiles with different air.


Agateophobia (approved):
Does what it says on the tin, good job.

PR #422, opened 2024-12-30
Makes so Mask hide the character name on the chat.

wgtjunior743:
## About The Pull Request

I always found strange how masks protects from exam but once you say something your name is exposed, and i would stay between is it right to know their person name even with the mask and etc.
Now every new vamp mask or animal mask hides the person name when they say, it does not hide from the admin logs.

Examples:

image

vamp_mask

animal_mask

## Why It's Good For The Game

Add another level of people being real secretive and distrust among peole who only uses masks, but there are methods against that, the masquerade contract will always reveal their name so it is pointless if you break the masquerade, also domination "WHO ARE YOU" will reveal their name.

## Changelog
:cl:
Tweak: Now vamp and animal masks hide their name on say
Fix: Head Nudist mask now hides your name on exam
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_say.dm
+++ b/code/modules/mob/living/carbon/human/human_say.dm
@@ -25,6 +25,30 @@
 			return ("Unknown")
 		else
 			return real_name
+	//Wod13 Masks
+	if(istype(wear_mask, /obj/item/clothing/mask/vampire))
+		var/obj/item/clothing/mask/vampire/V = wear_mask
+		if(V.voice_unknown)
+			return ("Unknown")
+		else
+			return real_name
+
+	if(istype(wear_mask, /obj/item/clothing/mask/animal))
+		var/obj/item/clothing/mask/animal/V = wear_mask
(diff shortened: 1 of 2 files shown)


GMR25:
Could it be possible to use the same name obfuscation as phones? Where the voice comes through as 'young adult man (204)' and similar, instead of merely 'Unknown'?

boy2mantwicethefam:
While masks usually aren't supposed to hide voices, there IS a discrepancy due to the fact that phones hide the user's name.

wgtjunior743:
> While masks usually aren't supposed to hide voices, there IS a discrepancy due to the fact that phones hide the user's name.

Indeed only if you added them as a contact It appears

wgtjunior743:
> The question here is to what level we want characters to recognise each other, or if we want them to at all. It wouldn't make much sense for all of our players to know each other's faces, but no one's voices. 

It Will even more difficult when thousand faces are a thing, i just find the anonymity lacking , even if you use a mask since day one , people will just know your name by talking once

PR #423, opened 2024-12-30
911 Dispatcher job for humans so we don't have to call personal phones in an emergency!

Decepteor:
## About The Pull Request

So this is a quick PR I've thrown together to add a new role for humans (and I suppose some ghouls can slip through the net). A dispatcher! 

image

There are a couple of things to document on this one, first is the code changes that mean that when 911 is dialed what is discussed (assuming a dispatcher picks up) is broadcast to all the police radios.

Obviously I've had to create the dispatch office too... which I've chosen the building attached to the PD for, it was the easiest choice but I don't mind iterating on this.

image

**Not part of this PR but:** Have we considered converting the map type away from the beyond default? Will help with merge conflicts!

## Why It's Good For The Game

Well I can only speak personally here but I think that having this role will help the human roles feel a tad more rounded, I've had a few moments of dissonance where I've wanted to call emergency services as a human and had basically no number I can call because the officers are all super busy getting murdered and the clinic staff are all vampires so they don't want to know.

Also I think it'd be a nice fun role too, not as stressful as playing police as you're hardly and active combatant but still involved with what's going on!

## Changelog
:cl:
add: Added the dispatcher role
add: Added the dispatch phone that is directly tied to the police radio as well
tweak: the building attached to the PD is now a dispatchers office.
/:cl:


Wildwestmars:
Can there be a way to hotline to the Clinic, since they have ambulances?
911 is also for medical emergencies.

Decepteor:
> Can there be a way to hotline to the Clinic, since they have ambulances? 911 is also for medical emergencies.

I've wired the calls into the new Clinic radio system.

githubuser4141:
It would be interesting if you could move NPC police around, and give them instructions to attack/detain/move somebody but that's probably out of scale for this PR. It's weird having there be like 5 or less cops for an entire modern city at any given moment, like IRL if you murder 5 cops you won't void the whole city of police. According to Google there's at least 8,000 officers in Los Angeles.

Decepteor:
> It would be interesting if you could move NPC police around, even spawn them if there's a lack of police, and give them instructions to attack/detain/move somebody but that's probably out of scale for this PR. It's weird having there be like 5 or less cops for an entire modern city at any given moment, like IRL if you murder 5 cops you won't void the whole city of police. According to Google there's at least 8,000 officers in Los Angeles.

I believe this PR https://github.com/WorldOfDarknessXIII/World-of-Darkness-13/pull/453 will be of interest to you @githubuser4141

Decepteor:
This PR is ready for review :)

code/modules/vtmb/jobs/dispatcher.dm
```diff
@@ -0,0 +1,38 @@
...
+	title = "Emergency Dispatcher"
+	faction = "Vampire"
+	total_positions = 2
+	spawn_positions = 2
+	supervisors = " the SF local government."
+	selection_color = "#7e7e7e"
+
+	outfit = /datum/outfit/job/dispatcher
+
+	access = list(ACCESS_HYDROPONICS, ACCESS_BAR, ACCESS_KITCHEN, ACCESS_ARMORY, ACCESS_MORGUE, ACCESS_WEAPONS, ACCESS_MINERAL_STOREROOM, ACCESS_THEATRE)
```
Miraviel, line 12: I believe they need none of these accesses declared (unless I'm missing something). We use these for lockers, currently, and with these, the dispatchers are able to open the armory closets + the NPCs will think of them as less of a threat when they hold a weapon.

code/modules/vtmb/jobs/dispatcher.dm
```diff
@@ -0,0 +1,38 @@
...
+	minimal_access = list(ACCESS_BAR, ACCESS_MINERAL_STOREROOM, ACCESS_THEATRE)
```
Miraviel, line 13: Same here

code/modules/vtmb/jobs/dispatcher.dm
```diff
@@ -0,0 +1,38 @@
...
+	paycheck_department = ACCOUNT_SRV
+	display_order = JOB_DISPLAY_ORDER_POLICE
+	exp_type_department = EXP_TYPE_POLICE
+
+	allowed_species = list("Ghoul", "Human")
+	species_slots = list("Ghoul" = 1)
+
+	duty = "Report emergencies to the correct emergency service."
+	minimal_masquerade = 0
+	my_contact_is_important = FALSE
```
Miraviel, line 24:
```suggestion
```

You can safely remove this, the parent is already FALSE.

Miraviel (changes requested):

code/modules/vtmb/jobs/dispatcher.dm
Decepteor: Should be resolved now :)

code/modules/vtmb/jobs/dispatcher.dm
Decepteor: Removed this too!

code/modules/vtmb/jobs/dispatcher.dm
Decepteor: And knocked this line out as well

Miraviel (approved):
I think we should give this a go, there will be probably small issue with it at first (as it is with every new job), but I cannot think of anything right now that would be a blocking issue.

code/modules/vtmb/electronics/phones/phone.dm
```diff
@@ -798,6 +798,39 @@
...
+	. = ..()
+	var/speaker = hearing_args[HEARING_SPEAKER]
+	if(speaker && online)
+		if(isliving(speaker) || istype(speaker, /obj/phonevoice) && get_dist(src, speaker) < 2)
+			var/message = hearing_args[HEARING_RAW_MESSAGE]
+			for(var/obj/item/police_radio/R in GLOB.police_radios)
+				R.dispatcher_talk(message)
+			for(var/obj/machinery/p25transceiver/P in GLOB.p25_tranceivers)
+				if(P.p25_network == "police" || P.p25_network == "clinic")
+					var/formatted = ""
```
Metekillot, line 827: There's no need to have two for loops here with an if statement. As well, we don't do single letter variables.

There's no need to loop through every single radio in the world. Refactor this to have the radios that would be listening to this frequency sorted in some way already instead of having to do the sorting every single time anyone anywhere sends a radio message.

code/modules/vtmb/electronics/phones/phone.dm
```diff
@@ -798,6 +798,39 @@
...
+	number = "911"
+	can_fold = 0
+	open_state = "redphone"
+	closed_state = "redphone"
+	folded_state = "redphone"
+
+/obj/item/vamp/phone/emergency/Initialize()
+	. = ..()
+	GLOB.phone_numbers_list += number
+	GLOB.phones_list += src
```
Metekillot, line 815:
```suggestion
	GLOB.phone_numbers_list += number //PSEUDO_M refactor global phone tracking
	GLOB.phones_list += src
```

Not a request for this PR but we need to refactor everything being tacked onto GLOB like this. There's also no need for us to have two fuckin global vars for phones...

code/modules/vtmb/electronics/phones/phone.dm
```diff
@@ -798,6 +798,39 @@
...
+			for(var/obj/item/police_radio/R in GLOB.police_radios)
+				R.dispatcher_talk(message)
+			for(var/obj/machinery/p25transceiver/P in GLOB.p25_tranceivers)
+				if(P.p25_network == "police" || P.p25_network == "clinic")
+					var/formatted = ""
+					if (isliving(speaker))
+						formatted = "[icon2html(src, world)]\[<b>DISPATCHER</b>\]: <span class='robot'>[message]</span>"
+					else
+						formatted = "[icon2html(src, world)]\[<b>CALLER</b>\]: <span class='robot'>[message]</span>"
+					P.broadcast_to_network(formatted, P.p25_network, 'sound/effects/radioclick.ogg', 10, FALSE)
```
Metekillot, line 832: Please check the validity of the dispatcher in a way that isn't them being living. We want to account for someone co-opting the dispatcher down the line.

Metekillot (changes requested):

code/modules/vtmb/electronics/phones/phone.dm
Decepteor: Oh as in make sure it's actually the dispatcher? I believe I can do that yes :)

code/modules/vtmb/electronics/phones/phone.dm
Decepteor: Well good news on the loop below it actually doesn't loop through all the radios, merely the transceivers of which there is one of each on the map! (I can remove the top loop because that's the old dispatch radio code, assuming we don't use it anymore)
@Metekillot

Decepteor: I will fix the one letter variables however.

Decepteor: (I do see a fix I can do to assign the transceivers too actually so that's easy enough!)

code/modules/vtmb/electronics/phones/phone.dm
Decepteor: Would you like me to commit this suggestion?

PR #424, opened 2024-12-30
Fixes Tremere Exploit

wgtjunior743:
## About The Pull Request

You can't use an empty blood bag for the runes that need it or use a teleport rune that is not yours.

## Why It's Good For The Game

Fix

## Changelog
:cl:
fix: Fix Tremere runes bugs/exploits.
/:cl:


lucifernix:
I think using someone else's teleport rune could be considered a feature to be honest. But the fix is really nice.

wgtjunior743:
> I think using someone else's teleport rune could be considered a feature to be honest. But the fix is really nice.

yeah but i am not sure in the lore if people without thaum can just use the power of runes

PR #442, opened 2025-01-02
Nerfs diablerie

faaaay:
## About The Pull Request

You no longer get reset to generation 13 by diablerie, as this is an extreme tool to allow (hypothetically) just about anyone to use on anyone they can robust. 
In addition, diablerie now simply lowers your generation by 1 on a successful attempt **(edit: when performed on someone of a lower generation),** or by 2 if the victim was at least 4 generations below you. _Failed_ diablerie raises your generation by 1.

## Why It's Good For The Game

The server is getting rid of many of the progress-loss features the codebase once had, and I think this is appropriate for that direction. The possibility of losing 10+ hours of XP after losing a fight might have suited what the original server was going for, but not this one, and it's not practical to effectively moderate every single use. 
This may even make diablerie more appealing, now that you're not causing a significant blow to an actual players's progress. Being evil is fun when characters suffer, not when _people_ lose a major chunk of XP through potentially no fault of their own.

## Changelog
:cl:
balance: Diablerie no longer affects the victim's generation. Failed diablerie raises the attacker's generation by 1.
balance: Diablerie will only lower your generation by 1, or by 2 if the victim is 4 or more generations below you.
/:cl:

### DIFF
--- a/code/modules/wod13/bloodsucking.dm
+++ b/code/modules/wod13/bloodsucking.dm
@@ -88,7 +88,7 @@
 					return
 
 				if(vse_taki)
-					to_chat(src, "<span class='userdanger'><b>YOU TRY TO COMMIT DIABLERIE OVER [mob].</b></span>")
+					to_chat(src, "<span class='userdanger'><b>YOU TRY TO COMMIT DIABLERIE ON [mob].</b></span>")
 				else
 					to_chat(src, "<span class='warning'>You find the idea of drinking your own <b>KIND</b> disgusting!</span>")
 					return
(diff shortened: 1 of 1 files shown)


YuiY1997:
This is such a needed change, especially in a game with the already controversial xp system that demands massive time investment

YuiY1997:
> I can just diablerize my metaclique friend till I get 7 gen, and then let them do the same, good PR

Administration problem, not code problem.

faaaay:
> I can just diablerize my metaclique friend till I get 7 gen, and then let them do the same, good PR

You have a chance to fail and go up a generation every time, which is higher if they're already gen 7. Alternatively, you could gain generations at about that pace by just earning XP at a high rate anyway.

Metekillot (approved):
sure, we're killing generation bullshit with hammers soon anyway and i'll take what i can get to lower the TDM crap

PR #454, opened 2025-01-04
Makes Curse an actual curse

AllCorruptingWyrm:
## About The Pull Request
In the context of roleplay, Clan Tremere is not a fun experience. They inherently lack any niche or the ability to be feared as they are in lore. Giovanni/Cappadocians/Kiasyd can summon ghosts and Tremere can as well, and Lasombra get more utility out of their shadow magics so with that said, to give the Camarilla something to make people fear Clan Tremere.

I have swapped damage types from fire (which can easily be bloodhealed I would add) to aggravated damage. The damage might need to be scaled down post-generation nerf but I think its a good start to let Tremere Do Their Thing.

## Why It's Good For The Game
Mentioned Above

## Changelog
:cl:
Makes curse deal cloneloss damage.
/:cl:

### DIFF
--- a/code/modules/vtmb/magic/pyramid.dm
+++ b/code/modules/vtmb/magic/pyramid.dm
@@ -361,7 +361,7 @@
 			cursed = namem
 			for(var/mob/living/carbon/human/H in GLOB.player_list)
 				if(H.real_name == cursed)
-					H.adjustCloneLoss(60)
+					H.adjustCloneLoss(25)
 					playsound(H.loc, 'code/modules/wod13/sounds/thaum.ogg', 50, FALSE)
 					to_chat(H, "<span class='warning'>You feel someone repeating your name from the shadows...</span>")
 					H.Stun(10)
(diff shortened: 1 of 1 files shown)


Plaguewalker:
This feels like a buff than just exchanging damages. Like 4 hits from this clonestate would pretty much nuke 240 health. IF MY MATH ISN'T WRONG

AllCorruptingWyrm:
> Changing it to cloneloss would be a big buff for a global ability. Cloneloss in general is extremely hard to heal outside of having specific medical items. To put it in perspective, blood heal removes -5 cloneloss per use which would mean someone would have to bloodheal 20 times (which would take 10 minutes) to recover from being cursed 4 times (100 cloneloss). Generation is not taken into account when healing cloneloss with bloodheal. The req for hearts may slow it down but it still has the potential to mess people up with no counter.
> 
> Outside of that, it functions codwise.

Yeah I'm aware.
I might just lower it to 15 instead. I think Trenere as they stand are a bit of a joke? And they really need something to scare people.

PR #397, opened 2024-12-27
[draft] nerfs potence

AllCorruptingWyrm:
## About The Pull Request

removes the knockdown from potence.

## Why It's Good For The Game

(insert multi para response as to why combat should not be trivalized)

## Changelog
Nerfs potence
;cl:

PR #461, opened 2025-01-05
Potence nerf and an unarmed combat change

boy2mantwicethefam:
Revival of #397 , this PR does two things:
1. It toggles the "gentle" variable to Potence's throw_impact, which disables damage and stuns taken from colliding with other things such as walls.
2. It reworks the punching knockdown system. Previously you needed Potence to have a chance to knock someone down (until higher levels of Potence would just be guaranteed to knock down and stunlock), now punching has a 10% chance per punch to knock someone down for 2 seconds, which increases up to 30% per punch depending on the difference between the puncher's physique and whichever is greater of the target's physique or dexterity. The difference for the stun chance increase has to be at least 2 (example: user's physique of 5 vs target's physique and dexterity of 3, which increases the stun chance to 20%). 

I don't like how unatomic the PR has become.

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -98,8 +98,6 @@ GLOBAL_LIST_EMPTY(selectable_races)
 	var/punchdamagelow = 1
 	///Highest possible punch damage this species can give.
 	var/punchdamagehigh = 10
-	///Damage at which punches from this race will stun
-	// var/punchstunthreshold = 20 //yes it should be to the attacked race but it's not useful that way even if it's logical
 	///Base electrocution coefficient.  Basically a multiplier for damage from electrocutions.
 	var/meleemod = 1
 	//For melee damage
@@ -1432,14 +1430,6 @@ GLOBAL_LIST_EMPTY(selectable_races)
 			target.apply_damage(damage, user.dna.species.attack_type, affecting, armor_block)
 			target.apply_damage(damage*1.5, STAMINA, affecting, armor_block)
 			log_combat(user, target, "punched")
-
-		///// WoD13 edit; punchstunthreshold variable on species was commented out.
-		// if((target.stat != DEAD) && damage >= user.dna.species.punchstunthreshold)
-		// 	target.visible_message("<span class='danger'>[user] knocks [target] down!</span>", "<span class='userdanger'>You're knocked down by [user]!</span>", "<span class='hear'>You hear aggressive shuffling followed by a loud thud!</span>", COMBAT_MESSAGE_RANGE, user)
-		// 	to_chat(user, "<span class='danger'>You knock [target] down!</span>")
-		// 	var/knockdown_duration = 40 + (target.getStaminaLoss() + (target.getBruteLoss()*0.5))*0.8 //50 total damage = 40 base stun + 40 stun modifier = 80 stun duration, which is the old base duration
-		// 	target.apply_effect(knockdown_duration, EFFECT_KNOCKDOWN, armor_block)
-		// 	log_combat(user, target, "got a stun punch with their previous punch")
 
 /datum/species/proc/spec_unarmedattacked(mob/living/carbon/human/user, mob/living/carbon/human/target)
 	return
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -1415,7 +1415,7 @@ GLOBAL_LIST_EMPTY(selectable_races)
 
 		if(user.potential >= 5)
 			var/atom/throw_target = get_edge_target_turf(target, user.dir)
-			target.throw_at(throw_target, rand(5, 7), 4, user)
+			target.throw_at(throw_target, rand(5, 7), 4, user, gentle = TRUE) //No stun nor impact damage from throwing people around
 
 		target.lastattacker = user.real_name
 		target.lastattackerckey = user.ckey
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -1430,6 +1430,17 @@ GLOBAL_LIST_EMPTY(selectable_races)
 			target.apply_damage(damage, user.dna.species.attack_type, affecting, armor_block)
 			target.apply_damage(damage*1.5, STAMINA, affecting, armor_block)
 			log_combat(user, target, "punched")
+		//Punches have a chance (by default 10%, up to 30%) to knock down a target for about 2 seconds depending on physique and dexterity.
+		//Checks if the target is already knocked down to prevent stunlocking.
+		if((target.stat != DEAD) && (!target.IsKnockdown()))
+			//Compare puncher's physique to the greater between the target's physique (robust enough to tank it) or dexterity (rolls with the punches)
+			var/modifier = clamp(user.get_total_physique() - max(target.get_total_physique(), target.get_total_dexterity()), 1, 3)
+			if(storyteller_roll(difficulty = 10 - modifier) == ROLL_SUCCESS)
+				target.visible_message("<span class='danger'>[user] knocks [target] down!</span>", "<span class='userdanger'>You're knocked down by [user]!</span>", "<span class='hear'>You hear aggressive shuffling followed by a loud thud!</span>", COMBAT_MESSAGE_RANGE, user)
+				to_chat(user, "<span class='danger'>You knock [target] down!</span>")
+				target.apply_effect(2 SECONDS, EFFECT_KNOCKDOWN, armor_block)
+				log_combat(user, target, "got a stun punch with their previous punch")
+
 
 /datum/species/proc/spec_unarmedattacked(mob/living/carbon/human/user, mob/living/carbon/human/target)
 	return
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -1435,8 +1435,7 @@ GLOBAL_LIST_EMPTY(selectable_races)
 
 		///// WoD13 edit; punchstunthreshold variable on species was commented out.
 		// if((target.stat != DEAD) && damage >= user.dna.species.punchstunthreshold)
-		// 	target.visible_message("<span class='danger'>[user] knocks [target] down!</span>", \
-		// 					"<span class='userdanger'>You're knocked down by [user]!</span>", "<span class='hear'>You hear aggressive shuffling followed by a loud thud!</span>", COMBAT_MESSAGE_RANGE, user)
+		// 	target.visible_message("<span class='danger'>[user] knocks [target] down!</span>", "<span class='userdanger'>You're knocked down by [user]!</span>", "<span class='hear'>You hear aggressive shuffling followed by a loud thud!</span>", COMBAT_MESSAGE_RANGE, user)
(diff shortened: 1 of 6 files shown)


Metekillot:
Why?

boy2mantwicethefam:
Previous PR reception + user asked for someone else to do it.

CatandWet:
Commenting out all that knockdown code has far more repercussions than just to fix potence, I really don't like it being removed since it basically turns unarmed into brute only slaps, instead of anything potent. Garou don't knockdown at all even at phys 5 with their unarmed until using something like ahroun's razor claws to boost damage high enough to hit the threshold for this knockdown you're removing and probably other stuff relies on it too.
I also don't think potence deserves being nerfed more when it already got made that if you don't run warrior/celerity you can't chain potence 5 punches very effectively; removing the stun will just knock someone back and have them immediately crawling and wiggling away which for everyone else who doesn't use it as a cheap smash someone against a wall until they die move, it will lose a lot of it's function and worth.
The way to 'fix' potence imo is just lessen the time you're on the ground; someone a while ago was going to make popping blood strength make you get up from knockdowns slightly quicker but that never happened sadly, despite also being a good alternative.

boy2mantwicethefam:
The Potence 5 stunlock is far worse than it is given credit for; the stun scales with the brute damage sustained by the mob and an initial hit at 20 damage (taking into account vampire vs vampire combat where brute damage is reduced, it's 40 damage against humans) will cause a 4.8-second stun (IIRC, it depends on whether the punch effect happens before or after the damage is applied), which only gets worse the more damage the target sustains at a rate of about existing brute damage * 0.4. There is no real counterplay to this other than turning into a simple-mob that can't be stunned (an ability only limited to a very low number of clans) or not getting hit in the first place.
I recall that the Garou have the advantage of automatically attacking the people they run into but I could be wrong.
The stun makes Potence an automatic "I win" button during combat because you only need to land one hit with it. This doesn't even touch its other function of multiplying melee damage (up to triple at max level), which nobody really engages with due to the aforementioned stun, people only know Potence for its stunlocks.

CatandWet:
"I recall that the Garou have the advantage of automatically attacking the people they run into but I could be wrong." They absolutely DO NOT have that, it is simple mob forms only that have the bump attacks which is arguably what makes those forms like protean/tzim warform etc. actually good.

I still think all that needs to be done is lower the time you're knocked down, stuns are nigh necessary in our fast paced combat where everyone can MMB jump around and celerity offscreen in a flash.

code/modules/mob/living/carbon/human/species.dm
```diff
@@ -99,7 +99,7 @@ GLOBAL_LIST_EMPTY(selectable_races)
 	///Highest possible punch damage this species can give.
 	var/punchdamagehigh = 10
 	///Damage at which punches from this race will stun
-	var/punchstunthreshold = 20 //yes it should be to the attacked race but it's not useful that way even if it's logical
+	// var/punchstunthreshold = 20 //yes it should be to the attacked race but it's not useful that way even if it's logical
```
Metekillot, line 102:
```suggestion
```
Don't leave commented out code inside of files. Either remove code, rework it, or leave it as is. We can track code changes with git already so there's no need for this, and it clutters up the codebase.

code/modules/mob/living/carbon/human/species.dm
```diff
@@ -1433,13 +1433,13 @@ GLOBAL_LIST_EMPTY(selectable_races)
...
-			target.apply_effect(knockdown_duration, EFFECT_KNOCKDOWN, armor_block)
-			log_combat(user, target, "got a stun punch with their previous punch")
+		///// WoD13 edit; punchstunthreshold variable on species was commented out.
+		// if((target.stat != DEAD) && damage >= user.dna.species.punchstunthreshold)
+		// 	target.visible_message("<span class='danger'>[user] knocks [target] down!</span>", "<span class='userdanger'>You're knocked down by [user]!</span>", "<span class='hear'>You hear aggressive shuffling followed by a loud thud!</span>", COMBAT_MESSAGE_RANGE, user)
+		// 	to_chat(user, "<span class='danger'>You knock [target] down!</span>")
+		// 	var/knockdown_duration = 40 + (target.getStaminaLoss() + (target.getBruteLoss()*0.5))*0.8 //50 total damage = 40 base stun + 40 stun modifier = 80 stun duration, which is the old base duration
+		// 	target.apply_effect(knockdown_duration, EFFECT_KNOCKDOWN, armor_block)
+		// 	log_combat(user, target, "got a stun punch with their previous punch")
```
Metekillot, line 1442:
```suggestion
```

code/modules/mob/living/carbon/human/species_types/golems.dm
```diff
@@ -12,7 +12,7 @@
 	siemens_coeff = 0
 	punchdamagelow = 5
 	punchdamagehigh = 14
-	punchstunthreshold = 11 //about 40% chance to stun
+	// punchstunthreshold = 11 //about 40% chance to stun
```
Metekillot, line 15:
```suggestion
```

code/modules/mob/living/carbon/human/species_types/golems.dm
```diff
@@ -168,7 +168,7 @@
 	name = "Silver Golem"
 	id = "silver golem"
 	fixed_mut_color = "ddd"
-	punchstunthreshold = 9 //60% chance, from 40%
+	// punchstunthreshold = 9 //60% chance, from 40%
```
Metekillot, line 171:
```suggestion
```

code/modules/mob/living/carbon/human/species_types/golems.dm
```diff
@@ -190,7 +190,7 @@
 	stunmod = 0.4
 	punchdamagelow = 12
 	punchdamagehigh = 21
-	punchstunthreshold = 18 //still 40% stun chance
+	// punchstunthreshold = 18 //still 40% stun chance
```
Metekillot, line 193:
```suggestion
```

code/modules/mob/living/carbon/human/species_types/golems.dm
```diff
@@ -328,7 +328,7 @@
 	brutemod = 0.5
 	punchdamagelow = 1
 	punchdamagehigh = 10
-	punchstunthreshold = 9
+	// punchstunthreshold = 9
```
Metekillot, line 331:
```suggestion
```

code/modules/mob/living/carbon/human/species_types/golems.dm
```diff
@@ -548,7 +548,7 @@
 	)
 	punchdamagelow = 0
 	punchdamagehigh = 1
-	punchstunthreshold = 2 //Harmless and can't stun
+	// punchstunthreshold = 2 //Harmless and can't stun
```
Metekillot, line 551:
```suggestion
```

code/modules/mob/living/carbon/human/species_types/golems.dm
```diff
@@ -697,7 +697,7 @@
 	burnmod = 2 // don't get burned
 	speedmod = 1 // not as heavy as stone
 	punchdamagelow = 4
-	punchstunthreshold = 7
+	// punchstunthreshold = 7
```
Metekillot, line 700:
```suggestion
```

code/modules/mob/living/carbon/human/species_types/golems.dm
```diff
@@ -901,7 +901,7 @@
 	heatmod = 2
 	speedmod = 1.5
 	punchdamagelow = 4
-	punchstunthreshold = 7
+	// punchstunthreshold = 7
```
Metekillot, line 904:
```suggestion
```

code/modules/mob/living/carbon/human/species_types/monkeys.dm
```diff
@@ -20,7 +20,7 @@
 	sexes = FALSE
 	punchdamagelow = 1
 	punchdamagehigh = 3
-	punchstunthreshold = 4 // no stun punches
+	// punchstunthreshold = 4 // no stun punches
```
Metekillot, line 23:
```suggestion
```

code/modules/mob/living/carbon/human/species_types/mushpeople.dm
```diff
@@ -16,7 +16,7 @@
 
 	punchdamagelow = 6
 	punchdamagehigh = 14
-	punchstunthreshold = 14 //about 44% chance to stun
+	// punchstunthreshold = 14 //about 44% chance to stun
```
Metekillot, line 19:
```suggestion
```

code/modules/mob/living/carbon/human/species_types/synths.dm
```diff
@@ -28,7 +28,7 @@
 	armor = 25
 	punchdamagelow = 10
 	punchdamagehigh = 19
-	punchstunthreshold = 14 //about 50% chance to stun
+	// punchstunthreshold = 14 //about 50% chance to stun
```
Metekillot, line 31:
```suggestion
```

code/modules/wod13/mannequin.dm
```diff
@@ -8,7 +8,7 @@
 	siemens_coeff = 0
 	punchdamagelow = 5
 	punchdamagehigh = 5
-	punchstunthreshold = 0 //about 40% chance to stun
+	// punchstunthreshold = 0 //about 40% chance to stun
```
Metekillot, line 11:
```suggestion
```

Metekillot (commented):
Regarding the total removal of potence punch stuns; that would be pending the current discipline rework, I believe? Otherwise that heavy of a balance change I'll want to double check with the rest of the team.

boy2mantwicethefam:
@CatandWet I don't think pretty much removing people's agency in combat is good for combat gameplay IMO. I'm not sure what fist-favoring builds World of Darkness as a setting has though, so I'd consider a buff to fists to make up for the loss of the stunlock.

Metekillot (approved):
at some point in the future i want us to rework throwing to be able to do damage without stunning, because supernaturals hurling each other through walls like dragon ball z sounds sick, but we're rocking with kiboshing the hard stuns

Metekillot:
Running some testing and if everything goes well we're good to merge this

CatandWet:
> Regarding the total removal of potence punch stuns; that would be pending the current discipline rework, I believe? Otherwise that heavy of a balance change I'll want to double check with the rest of the team.

As someone who plays a lot I think this is a very bad idea, where is the discussion?
This does not just affect potence punches this meddles with unarmed in it's entirety.

Metekillot:
I'd noticed some weirdness in testing and had actually been spending the last week working on getting the SpacemanDMM debugger working. Discord is not a good place for discussions on balance changes like this due to the immense scroll and tendency for users to tower post.

Metekillot (changes requested):
After some testing, I'm not on board with totally removing knockdown from unarmed combat. Instead, please add a check that benefits from the higher of your dexterity or physique to determine if you get knocked down by a punch. Please include some amount of randomness in it. @TheCarnalest does this seem reasonable to you?

Keeping potence throws as gentle is fine with me.

Metekillot:
Specifically I'm requesting that knockdowns not be taken out of combat and
instead be something that you can resist with a dice roll and your stats.

Metekillot (approved):

Metekillot (approved):

PR #478, opened 2025-01-06
Jaegerspas-XV added

cornercube:
## About The Pull Request
image

adds the Jaegerspas-XV (SPAS-15)
semi-automatic, magazine-fed, shotgun
in exchange for slightly less damage (-10%), it can be tactically reloaded and fired without pumping

fiddles with shotgun shell stun
slugs increased in stun from 20 -> 25
individual pellets on buck decreased from 20 -> 4

## Why It's Good For The Game
VTMB gun
an upgrade over the pump-action

auto-shottie will allow stunlocking too easily
each individual shot pellet dealt 20 stun, causing 160 stun if all pellets hit
this would be abuseable with a semi-auto shotgun, so its reduced to 4 stun per pellet, for 32 stun if all of them hit, to make it less of a tractor beam of death, so hitting with all rounds is still longer than with a slug, but length of stun is variable depending on how many pellets hit (and therefore range)
regular slug slightly increased to 25 stun, since 20 stun wasnt actually that long

## Changelog
:cl:
add: the jaegerspas-XV and its sprites
balance: buck deals way less stun, slug deals slightly more stun
change: shot pellets in their own path instead of a childe of the slug
/:cl:


lucifernix:
There are merge conflicts.

boy2mantwicethefam:
Do stuns really add up like that?

Agateophobia (approved):
No reason not to merge it, even if it is utterly busted, it's not added anywhere and gun bloat good. Merged.

PR #484, opened 2025-01-06
Undoes PR #475

theXDkidoflol:
## About The Pull Request
Undoes the effects of https://github.com/WorldOfDarknessXIII/World-of-Darkness-13/pull/475, which had the sole change of removing coordinates from the Dispatch radio.

## Why It's Good For The Game
So, for an explanation, some context should be provided.

At the time PR 475 was made, it was during a round in which the Giovanni Mansion was raided and the entire P.D plus a follow-up SWAT team was killed. The raid was based on the Dispatch Radio reporting multiple murders at the Giovanni mansion. Before the round had even finished, PR 475 was made and speedmerged.

475 does not resolve the issue at all, and solely harms intended usage of the Dispatch Radio.

The intended usage is presumably to make it so any crimes in public can be responded to in a timely matter, as though one of the NPCs called the police. This is not practically possible with PR 475 merged, as external area maps cover HUGE zones; Fisherman's Wharf alone covers nearly an entire side of the city, meaning any murder reported there is nigh-impossible to find.

Instead, now it can ONLY be used to gain warrants. When five murders are consecutively reported in an area like the Giovanni Mansion, you now have no other use for the Dispatch radio than to investigate it. Without coordinates, it becomes infinitely easier to focus on these 'restricted area' incidents, too. Plus, it can give reason to search an ENTIRE premises instead of just going for a specific location (i.e raiding the underground of the bar when the body was just upstairs in the showers).

A better solution might be to make it so that at the time a crime is committed, it checks if there's a living NPC within view range. I don't have the skill to add that, though. As it stands, though, 475 not only isn't a solution, it harms intended usage.

## Changelog
:cl:
balance: Dispatch radio has coordinates again.
/:cl:

### DIFF
--- a/code/modules/wod13/witness.dm
+++ b/code/modules/wod13/witness.dm
@@ -41,15 +41,15 @@
 			if(last_shooting+50 < world.time)
 				last_shooting = world.time
 				var/area/A = get_area(location)
-				say("Citizens report hearing gunshots at [A.name], to the [get_cardinal_direction(location.x, location.y)]...")
+				say("Citizens report hearing gunshots at [A.name],[get_cardinal_direction(location.x, location.y)], [location.x]:[location.y]...")
 		if("victim")
 			if(last_shooting_victims+50 < world.time)
 				last_shooting_victims = world.time
 				var/area/A = get_area(location)
-				say("Active firefight in progress at [A.name], wounded civilians, to the [get_cardinal_direction(location.x, location.y)]...")
+				say("Active firefight in progress at [A.name], wounded civilians,[get_cardinal_direction(location.x, location.y)], [location.x]:[location.y]...")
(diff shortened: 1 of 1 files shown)


xSlumLord:
The ideal solution is to change the dispatcher radio to not be global. It shouldn't be reporting crimes that nobody can see or report. It should be kept to witnessed crimes by NPCs and gunshots in locations where they could be heard. Crimes in the sewers should also go unreported.

Metekillot:
I'd recommend having a mediated discussion for PRs that serve solely to revert a PR that you, yourself, didn't create. It can come across as quite adversarial

theXDkidoflol:
alright, lucifer's claimed he'll implement a better fix shortly

PR #281, opened 2024-12-04
Generation + Blood + Discipline Rework

TheCarnalest:
## About The Pull Request
Completely reworks a whole bunch of systems.

### Generation
- [x] Make maximum bloodpool conform to tabletop
- [x] Implement "spending multiple blood points per turn"

### Blood
- [x] Change bloodpools to be tabletop-accurate (all NPCs have a max bloodpool of 10)
- [x] Completely refactor bloodpool and blood_volume
- [x] Replace direct bloodpool and blood_volume setting with new procs
- [x] Implement elders having more potent/concentrated blood

### Disciplines
- [x] Make Disciplines easily accessible outside of themselves
- [x] Split Discipline powers from Disciplines themselves
- [x] Cancelable Disciplines + make Potence, Celerity, and Fortitude work properly
- [x] Completely refactor Disciplines to be sane

## Why It's Good For The Game
BETTER CODE ALL AROUND, BETTER BLOOD STUFF BALANCE, BETTER DISCIPLINES

## Changelog
:cl:
add: Unimaginably long changelog I'll need to do later
/:cl:

PR #485, opened 2025-01-07
Daimonion Rework (Baali Discipline)

Nokia2Student:
## About The Pull Request

Reworks Daimonion to make sure it follows the lore.

•  Sense the Sin - Baali stares into target's soul, which exposes theirs greatest weaknesses and weakpoints.

•• Fear of the Void Below - Baali sends the target into uncontrollable frenzy, paralyzing them with fear.

••• Conflagration - Baali unleashes infernal fireball, which on contact deals aggravated damage as well as sets the target on fire.

•••• Psychomachia - Baali summons invisible to others but target creature that tries to catch them, if catched it paralyzes them and deals illusionary wounds.

••••• Condemnation - Baali sacrifices their own vitae in order to place a curse on the target. Beware, the stronger the curse the greater the price.

## Why It's Good For The Game

Canonizes the Baali Discipline - Daimonion. A lot of Baali players complained that Daimonion wasn't the same as in lore, that fixes it. The changes allow stealthy approach for Baali. Instead of a mix of different abilities (Protean, Vicissitude) now they'll have their own!

## Changelog
:cl:
add: Reworked the dots, added a bunch of new code
balance: Buffed Daimonion 3 by adding Aggravated Damage to it, as it should be, as well as painted it black (Buffed style points).
/:cl:


code/modules/mob/living/simple_animal/hostile/hostile.dm
```diff
@@ -85,6 +85,7 @@
 		if(CheckEyewitness(src, src, 5, FALSE))
 			SEND_SOUND(src, sound('code/modules/wod13/sounds/masquerade_violation.ogg', 0, 0, 75))
 			to_chat(src, "<span class='userdanger'><b>MASQUERADE VIOLATION</b></span>")
+			spawn(5 SECONDS)
```
Metekillot, line 88: Why did you create a spawn here?

code/modules/projectiles/projectile/magic.dm
```diff
@@ -681,7 +681,14 @@
 	R.color = color
 
 /obj/projectile/magic/aoe/fireball/baali
-	color = "#2dff00"
+	color = "#1c1f1d"
```
Metekillot, line 684: you could use rgb here; there should be a built-in extension that would let you adjust the color on sight

code/modules/projectiles/projectile/magic.dm
```diff
@@ -681,7 +681,14 @@
...
+	damage = 25
+	damage_type = CLONE
+
+/obj/projectile/magic/aoe/fireball/baali/on_hit(target)
+	. = ..()
+	var/obj/effect/fire/R = new(get_turf(target)) // Creates fire object on the hit if successfully landed
```
Metekillot, line 690:
```suggestion
	var/obj/effect/fire/created_fire = new(get_turf(target)) // Creates fire object on the hit if successfully landed
```

code/modules/projectiles/projectile/magic.dm
```diff
@@ -681,7 +681,14 @@
...
+	R.color = color
```
Metekillot, line 691:
```suggestion
	created_fire.color = src.color
```

code/modules/vtmb/disciplines.dm
```diff
@@ -1653,53 +1655,263 @@
 	desc = "Get a help from the Hell creatures, resist THE FIRE, transform into an imp. Violates Masquerade."
 	icon_state = "daimonion"
 	cost = 1
-	ranged = FALSE
-	delay = 150
-	violates_masquerade = TRUE
+	delay = 100
+	violates_masquerade = FALSE
```
Metekillot, line 1659:
```suggestion
	violates_masquerade = TRUE
```
Why did you set this to FALSE?

code/modules/vtmb/disciplines.dm
```diff
@@ -1653,53 +1655,263 @@
...
+		if(K.get_discipline("Melpominee") && !(H.clane?.name == "Daughters of Cacophony"))
+			to_chat(caster, "They fear that the fact they stole Daughters of Cacophony's Melpominee will be known.")
+		if(K.get_discipline("Daimonion") && !(H.clane?.name == "Baali"))
+			to_chat(caster, "They fear that the fact they stole Baali's Daimonion will be known.")
+		if(K.get_discipline("Temporis") && !(H.clane?.name == "True Brujah"))
+			to_chat(caster, "They fear that the fact they stole True Brujah's Temporis will be known.")
+		if(K.get_discipline("Valeren") && !(H.clane?.name == "Salubri"))
+			to_chat(caster, "They fear that the fact they stole Salubri's Valeren will be known.")
+		if(K.get_discipline("Mytherceria") && !(H.clane?.name == "Kiasyd"))
+			to_chat(caster, "They fear that the fact they stole Kiasyd's Mytherceria will be known.")
```
Metekillot, line 1802: Two things:

This needs to be handled by status_effects, which already have a processing subsystem and many variables that are optimized for handling things like this.

All of these messages need to be stored somewhere and referenced, there is no reason why you need dozens of lines just for messages in the middle of the disciplines file.

code/modules/vtmb/disciplines.dm
```diff
@@ -1653,53 +1655,263 @@
...
-			caster.dna.species.GiveSpeciesFlight(caster)
-			spawn(delay+caster.discipline_time_plus)
-				if(caster)
-					caster.dna.species.RemoveSpeciesFlight(caster)
-					caster.playsound_local(caster.loc, 'code/modules/wod13/sounds/protean_deactivate.ogg', 50, FALSE)
+			var/mypower = caster.get_total_social()
+			var/theirpower = target.get_total_mentality()
+			if(theirpower >= mypower)
+				to_chat(caster, "<span class='warning'>[target] has too much willpower to induce fear into them!</span>")
+				return
```
Metekillot: This could really benefit from dice rolling instead of being a flat if then, but the variable names are good.

code/modules/vtmb/disciplines.dm
```diff
@@ -1653,53 +1655,263 @@
...
+			to_chat(target, "<span class='warning'><b>Your mind is enveloped by your greatest fear!</span></b>")
+			var/mob/living/carbon/human/H = target
+			if(!H.in_frenzy) // Cause target to frenzy
+				H.enter_frenzymod()
+				H.Paralyze(3 SECONDS)
+				addtimer(CALLBACK(H, TYPE_PROC_REF(/mob/living/carbon, exit_frenzymod)), 3 SECONDS)
```
Metekillot, line 1843:
```suggestion
			to_chat(target, "<span class='warning'><b>Your mind is enveloped by your greatest fear!</span></b>")
			if(!target.in_frenzy) // Cause target to frenzy
				target.enter_frenzymod()
				target.Paralyze(3 SECONDS)
				addtimer(CALLBACK(target, TYPE_PROC_REF(/mob/living/carbon, exit_frenzymod)), 3 SECONDS)
```
just type the target in the args of the proc definition

code/modules/vtmb/disciplines.dm
```diff
@@ -1653,53 +1655,263 @@
...
-			caster.drop_all_held_items()
-			BAT.Shapeshift(caster)
-			spawn(delay+caster.discipline_time_plus)
-				if(caster && caster.stat != DEAD)
-					BAT.Restore(BAT.myshape)
-					caster.Stun(15)
-					caster.do_jitter_animation(30)
-					caster.playsound_local(caster.loc, 'code/modules/wod13/sounds/protean_deactivate.ogg', 50, FALSE)
+			var/turf/start = get_turf(caster)
+			var/obj/projectile/magic/aoe/fireball/baali/H = new(start)
```
Metekillot, line 1846: are you seriously using the same single letter variable for two entirely different types?

code/modules/wod13/onyxcombat.dm
```diff
@@ -736,7 +736,7 @@
 /mob/living/proc/update_blood_hud()
 	if(!client || !hud_used)
 		return
-	maxbloodpool = 10+((13-generation)*3)
+	maxbloodpool = (10+((13-generation)*3))-cursed_bloodpool
```
Metekillot, line 739: we need to come up with a proc for calculating this

code/modules/projectiles/projectile/magic.dm
Nokia2Student: VS Code allows me to change that already
image

code/modules/mob/living/simple_animal/hostile/hostile.dm
Nokia2Student: Originally it was added to fix an issue with simplemobs causing a masquerade breach each 2 seconds. The thing that it was used for was scrapped away so I might as well remove it and let others fix it

code/modules/vtmb/disciplines.dm
Nokia2Student: Because abilities like Daimonion 1, 2, 4, 5 aren't supposed to be masquerade breaching. There's no way to make that value separate for each discipline.

code/modules/vtmb/disciplines.dm
Nokia2Student: It checks for ment/social and generation, if any of it aren't above the target's then it just cancels the discipline.

code/modules/vtmb/disciplines.dm
Nokia2Student: is_frenzy accepts mob/living/carbon/human only. target is mob/living

code/modules/wod13/onyxcombat.dm
Nokia2Student: why? There's already a defined variable in mob/living

code/modules/vtmb/disciplines.dm
Nokia2Student: they are already stored in a proc that was created exactly for the purpose for printing out messages. Tell me more about status_effects, though, and how would it make the code better?

code/modules/vtmb/disciplines.dm
Nokia2Student: fixed

code/modules/mob/living/simple_animal/hostile/hostile.dm
Nokia2Student: fixed

code/modules/vtmb/disciplines.dm
Metekillot: > is_frenzy accepts mob/living/carbon/human only. target is mob/living

i am saying type target to be mob/living/carbon/human

code/modules/vtmb/disciplines.dm
Metekillot: I'm saying add a little bit of randomness into it so that it isn't the same result every time if someone figures out they have better/worse stats than someone else.

code/modules/vtmb/disciplines.dm
Metekillot: They shouldn't be smack dab in the middle of a file, and they should be stored in their own variable due to the amount of them.

Regarding status_effects: search the code
/datum/status_effects

code/modules/vtmb/disciplines.dm
Metekillot: > Because abilities like Daimonion 1, 2, 4, 5 aren't supposed to be masquerade breaching. There's no way to make that value separate for each discipline.

You can actually subtype the discipline itself for this.

code/modules/vtmb/disciplines.dm
Nokia2Student: if you set it this way then you can't compile it. byond complains about possible loops.

code/modules/vtmb/disciplines.dm
Nokia2Student: they can always figure out if they have better/worse stats by using mental disciplines, but adding randomness factor would be good.

code/modules/vtmb/disciplines.dm
Nokia2Student: there's no /datum/status_effects in the code.

Nokia2Student: > They shouldn't be smack dab in the middle of a file, and they should be stored in their own variable due to the amount of them.

Do you want me to store clan names and disciplines in the variable?

code/modules/vtmb/disciplines.dm
Nokia2Student: > You can actually subtype the discipline itself for this.

I'm not quite sure what you mean by that. Can I have an example?

code/modules/vtmb/disciplines.dm
wgtjunior743: that is how the entire code of the game works for now, it always a check of social vs mentality and the chance is static, there is a refactor of disciplines going on which will change multiple stuff, then it will be better to balance these mental disciplines

code/modules/vtmb/disciplines.dm
wgtjunior743: > > You can actually subtype the discipline itself for this.
> 
> I'm not quite sure what you mean by that. Can I have an example?

You could try to do like this: https://github.com/WorldOfDarknessXIII/World-of-Darkness-13/pull/281/files
But it is pointless at the moment just have it break or not the masquerede for now, also it is important that you know: neither me or Metekillot has the final say or anything here, only the mainteners do so if youwould  like you could keep the code the way it is wait until a maintener comes, if they see and it and think that it is alright it will be merged as it is .

PR #497, opened 2025-01-08
Updates About Me for Garou

faaaay:
## About The Pull Request

The About Me button is outdated for werewolves, and doesn't show Lockpicking/Athletics or the stat bonuses of archetypes. Likely due to the added complexity of the Garou species, they additionally would not be given their own ATM PIN, effectively making them completely broke. This fixes both.

## Why It's Good For The Game

The update to displaying stats is just a plain oversight fix.
The latter fix makes bank_id and account_id variables of mob/living/carbon and not just carbon/human, which I don't _believe_ will cause any issues. I didn't see any runtimes when spawning.

## Changelog
:cl:
fix: updated the About Me verb for Garou to correctly display your stats
fix: Garou are now correctly assigned a bank account PIN
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon_defines.dm
+++ b/code/modules/mob/living/carbon/carbon_defines.dm
@@ -152,4 +152,7 @@
 	var/diablerist = FALSE
 	var/antifrenzy = FALSE
 
+	var/account_id //Making bank accounts work for Garou is a pain if this isn't defined here
+	var/bank_id
+
 	COOLDOWN_DECLARE(bleeding_message_cd)
--- a/code/modules/mob/living/carbon/human/human_defines.dm
+++ b/code/modules/mob/living/carbon/human/human_defines.dm
@@ -64,8 +64,6 @@
 	var/static/list/can_ride_typecache = typecacheof(list(/mob/living/carbon/human, /mob/living/simple_animal/slime, /mob/living/simple_animal/parrot))
 	var/lastpuke = 0
 	var/last_fire_update
-	var/account_id
-	var/bank_id
 
 	var/hardcore_survival_score = 0
 	/// For agendered spessmen, which body type to use
--- a/code/modules/vtmb/werewolf_species.dm
+++ b/code/modules/vtmb/werewolf_species.dm
@@ -62,11 +62,13 @@
 				if(A.objectives)
 					dat += "[printobjectives(A.objectives)]<BR>"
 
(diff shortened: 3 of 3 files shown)


faaaay:
> I hope this gets merged soon, there's some iffy behaviour with garou form changing and stats not being reset/updated properly that I want to fix but this likely needs merging before that. I'm no code reviewer but peeking at the code it's just updating garou's about me to be on par with other species', looks good to me.

I'm fairly sure this is caught up in the code freeze, and I didn't get around to making it compatible with that yet. It's not unlikely that when the code freeze ends, this issue will have been fixed.

Wildwestmars:
It'd be silly if this doesn't get merged because of the code freeze because this is a bug fix for a rather serious issue.

code/modules/mob/living/carbon/carbon_defines.dm
```diff
@@ -152,4 +152,7 @@
 	var/diablerist = FALSE
 	var/antifrenzy = FALSE
 
+	var/account_id //Making bank accounts work for Garou is a pain if this isn't defined here
```
Metekillot, line 155: This would be simplified if you just attached it to the mind datum instead of sticking it on a given mob. I suggest doing that instead because the mind also doesn't interface with species or mob type.

code/modules/mob/living/carbon/human/human_defines.dm
```diff
@@ -64,8 +64,6 @@
 	var/static/list/can_ride_typecache = typecacheof(list(/mob/living/carbon/human, /mob/living/simple_animal/slime, /mob/living/simple_animal/parrot))
 	var/lastpuke = 0
 	var/last_fire_update
-	var/account_id
-	var/bank_id
```
Metekillot, line 68: You actually making the right decision with pulling this shit off of the human type

Metekillot (changes requested):
Move the about me information on to the mind for the garou, and if you want to do everyone a favor you could do it for everyone else too but if you framework something for garou we could probably do it for all the rest of them and get rid of the headache

Metekillot:
I have a complete rework of bank accounts coming out in the next week ish, inspired by this PR. However, as a result, I'll be totally tearing apart all the work you've done already. My apologies.

faaaay:
> I have a complete rework of bank accounts coming out in the next week ish, inspired by this PR. However, as a result, I'll be totally tearing apart all the work you've done already. My apologies.

Not an issue, I'm all for the rework.

PR #506, opened 2025-01-09
The crafting UI will now update as soon as an item finishes crafting

boy2mantwicethefam:
Small change. No more random deciseconds of extra time until the UI updates. Most noticeable with low-crafting-time items like sorting weed.


Agateophobia (approved):
Works, approved and loved. Merge time.

PR #507, opened 2025-01-09
Injury slowdown will now scale with the user's max health

boy2mantwicethefam:
This PR makes the movement slowdown scale with the user's max health, so that relatively minor injuries no longer slow down as hard as they should.
Obvious balance implications in the way that high-physique characters and lower-generation vampires can move during combat, however.
This also fixes a potential issue where high-health characters could be slowed down far more than lower-health characters because of how the slowdown calculation worked, which didn't scale with max health.

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -1262,9 +1262,11 @@
 		remove_movespeed_modifier(/datum/movespeed_modifier/damage_slowdown_flying)
 		return
 	var/health_deficiency = max((maxHealth - health), staminaloss)
-	if(health_deficiency >= 40)
-		add_or_update_variable_movespeed_modifier(/datum/movespeed_modifier/damage_slowdown, TRUE, multiplicative_slowdown = health_deficiency / 75)
-		add_or_update_variable_movespeed_modifier(/datum/movespeed_modifier/damage_slowdown_flying, TRUE, multiplicative_slowdown = health_deficiency / 25)
+	var/health_scale_modifier = maxHealth/100 //If the human has a higher amount of max health this will be reflected in the amount of damage required to slow down.
+	var/health_slowdown = health_deficiency / (75 * health_scale_modifier)
+	if(health_deficiency >= (40 * health_scale_modifier))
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1262,9 +1262,11 @@
 		remove_movespeed_modifier(/datum/movespeed_modifier/damage_slowdown_flying)
 		return
 	var/health_deficiency = max((maxHealth - health), staminaloss)
-	if(health_deficiency >= 40)
-		add_or_update_variable_movespeed_modifier(/datum/movespeed_modifier/damage_slowdown, TRUE, multiplicative_slowdown = health_deficiency / 75)
-		add_or_update_variable_movespeed_modifier(/datum/movespeed_modifier/damage_slowdown_flying, TRUE, multiplicative_slowdown = health_deficiency / 25)
+	var/health_scale_modifier = maxHealth/100 //If the human has a higher amount of max health this will be reflected in the amount of damage required to slow down.
```
Metekillot, line 1265: i'd recommend making this calculation based off a GLOBAL_VAR instead of arbitrarily /100, so admins or whoever can adjust it on the fly in the future or just to facilitate debugging.

Metekillot (commented):
good to see clearly labeled variables that self-document their purpose. please try to make your commit messages more informative though, so people scrolling through the commits can see who did what, where, and why.

boy2mantwicethefam:
I've been doing "clearly labeled variables" for a long time, I only tend to use stuff like H, M and so on when it comes to the previously mentioned mob types.

code/modules/mob/living/carbon/human/human.dm
boy2mantwicethefam: It's just a way to scale the damage slowdown in order to allow higher-health characters to be proportionally affected (and if someone somehow has a lower amount of max health it will also scale off that), this shouldn't get out of the proc. The only thing I'd like to do is turn the "100" magic number into a proper global define that determines the base human max health but that's out of the scope of this PR, something like that shouldn't be bundled with a balance PR.

Metekillot (approved):

PR #510, opened 2025-01-10
Make Dice Rolls Pretty

hex37:
## About The Pull Request
- Adds a version of the storyteller roll that outputs the result to the player
- Adds game preference for toggling storyteller roll reports
- Adds Preference tab option for toggling storyteller roll reports
- Adds admin command for rolling cute lil dice at yourself (Turn on storyteller roll reporting first!)

x

## Why It's Good For The Game
- Gives a visual aid for players who want to know the odds of their rolls
- Debugging for a future when we actually use storyteller rolls instead of what we're currently doing


Metekillot (changes requested):
You gotta get rid of the use of 'usr' my man. And please move the rolls onto /mob instead of them being a global proc. Good work so far otherwise

hex37:
@Metekillot It's a fair complaint- are you willing to meet in the middle and have the roller default to usr if no target is given? That way we don't have to redo all the existing rolls and conflict with change-freeze code.

Metekillot:
> @Metekillot It's a fair complaint- are you willing to meet in the middle and have the roller default to usr if no target is given? That way we don't have to redo all the existing rolls and conflict with change-freeze code.
> 
> 

Once you update the code to meet standards I'll add to it to redo the existing rolls

hex37:
> > @Metekillot It's a fair complaint- are you willing to meet in the middle and have the roller default to usr if no target is given? That way we don't have to redo all the existing rolls and conflict with change-freeze code.
> 
> Once you update the code to meet standards I'll add to it to redo the existing rolls

Alright- I nixed every single usr var, save the one in the admin verb since it's just for debugging

Metekillot (approved):

PR #512, opened 2025-01-10
Fix ghoul Discipline bug

TheCarnalest:
## About The Pull Request
Ghouls were locked out of all unique Disciplines due to an oversight when locking whitelisted Disciplines to people with that whitelist.

## Why It's Good For The Game
Let ghouls have fun for once.

## Changelog
:cl:
fix: Ghouls can have unique Disciplines again.
/:cl:


lucifernix (approved):
These two lines of code look good to me.

PR #514, opened 2025-01-10
Gives CQC to Police Sergeants

SELFHELL:
## About The Pull Request

Pretty self-explanatory, but this brings down the CQC Martial Art for Police Sergeants too.

## Why It's Good For The Game

Gives the PD more people able to actually nonlethally restrain people, as compared to being forced to use the shotgun 4/5 times. when their victim resists.

## Changelog
:cl:
ad Gives CQC to Police Sergeants
/:cl:

### DIFF
--- a/code/modules/vtmb/jobs/police.dm
+++ b/code/modules/vtmb/jobs/police.dm
@@ -132,4 +132,6 @@
 
 /datum/outfit/job/police_sergeant/post_equip(mob/living/carbon/human/H)
 	..()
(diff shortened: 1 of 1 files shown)


Metekillot:
An uncounterable three click knockout doesn't seem like the logical leap for a lack of non lethal tools.

SELFHELL:
Fair enough,

Metekillot:
I respect the hustle on adding more tools for police to be an actual... policing force. But ideally, even if the force is overwhelming, it's something you can still flee from.

PR #516, opened 2025-01-11
Readjusts values & lessens the amount of corpsed bodies

AllCorruptingWyrm:
## About The Pull Request

The following pr makes it so the total cost of one NPC is roughly 2K if you trade in every single organ (Going by the cost = value). This PR also significantly buffs drug selling and weed trading as well! Encouraging players to set up farms and defend them as a means of establishing territory.

## Why It's Good For The Game

As mentioned above.

## Changelog
:cl:
Lowers the value of all organs to equal 2K.
Increases the value of weed to 350, meth to 650, and cocaine to 750.
/:cl:


Metekillot (approved):
Encouraging people to sell drugs and interact with each other manufacturing those drugs is much better to me than encouraging them to murderhobo NPCs

PR #487, opened 2025-01-07
The Rolening - Camarilla and Anarchs expanded!

AllCorruptingWyrm:
## About The Pull Request

This PR adds Hounds as a replacement for the unwashed horde of Scourge scum, and adds a generic Anarch role for the anarchs. It also adds Keepers and Harpies (Elysium security/Diplomat types) to the Camarilla and they basically are just copied over Seneschals with more of a focus. Hounds are implemented to fit in for Scourge which has been made a singular job slot. Hounds are either helping the Sheriff or Scourge in their job too. Hounds also lose their badge so they cant slug around their MP5s without just cause as we live in a post-911 world, people.

Emissary is made 2nd in Command, Reeves (Anarch sheriff expy) are added as well. 
image
Hound loadout
image
Anarch

image
Roles in game

## Why It's Good For The Game

Basically mentioned above; This was done to alleviate tdm and make the Camarilla less interested in mogging out really.

## Changelog
:cl:
Adds Reeves (Anarch Sheriffs)
Adds Keepers/Harpies (An elysium maintainer and diplomat role)
Makes Scourge a 1 slot role, you are the assassin of the Prince.
Adds Hounds, which act as the footsoldiers of the Camarilla
Adds a generic "Anarch" role which work in the Bar, also a foot soldier role too.
/:cl:

PR #517, opened 2025-01-11
Turns vampire_faction strings into FACTION_[faction] defines

mrhugo13:
## About The Pull Request
Turns a bunch of faction strings & string checks like "Camarilla" and "Anarchs" and "Sabbat" into defines for the sake of practicality & standardisation

Will likely conflict with a few ongoing PRs like #487 and #493 albeit should be fairly easy conflicts to resolve, even if this PR gets merged first the other PRs linked should still work about the same since all this does is moving these string checks to defines of the same nature.

Compiles & Doesn't runtime (atleast the relevant code changed doesn't, the four runtimes pictured have nothing to do with any of the files i've changed.)

## Why It's Good For The Game
Better, more standardised, code that will improve the development pipeline overall

## Changelog
Unnecessary for a completely code-/development-only change.


mrhugo13:
Will have to have a second go at this with var/faction as i presumed there only was var/vampire_faction to deal with

mrhugo13:
> Will have to have a second go at this in a second with var/faction as i presumed there only was var/vampire_faction to deal with

Nevermind to this, var/faction has more to do with simplemobs than sect stuff which is what this PR is primarily concerned with.

Metekillot:
At a glance (haven't looked at the code) what we want to do is have factions tracked on datums, actually, rather than checking against a given faction "name". If we have them tracked by a singleton datum, we don't have to tiptoe on eggshells changing faction names or what-have-you.

Metekillot:
You are on the right track with making defines for things like this, defines DO make the code safer, and when I'm able to review the code it will likely be a better solution than the naked strings we're using now. I just wanted to explain the intent with moving toward datums and signals handling most of this

mrhugo13:
> You are on the right track with making defines for things like this, defines DO make the code safer, and when I'm able to review the code it will likely be a better solution than the naked strings we're using now. I just wanted to explain the intent with moving toward datums and signals handling most of this

Thanks! I'll be converting this PR to a draft in the meantime while i explore the bitflag/bitfield/datum-ising path you mentioned.

mrhugo13:
New commit causes a rather massive amount of runtimes and is still WIP but i don't have enough time for it today so it will have to do until next i get to this, hopefully tomorrow and every day until it eventually works flawlessly.

*cough*

Metekillot (approved):

Metekillot:
approved pending a rollback to the finished string defines work; we'll work together on the signal bits because our implementation is so much older than tg's that we'll wanna make sure that everything is copacetic

mrhugo13:
> approved pending a rollback to the finished string defines work; we'll work together on the signal bits because our implementation is so much older than tg's that we'll wanna make sure that everything is copacetic

Done!

PR #519, opened 2025-01-11
Auspex ghosts should no longer be polled for midround roles

boy2mantwicethefam:
This includes Question To Ancestors, summoning demons and creating gargoyles.
Why? Because Auspex ghosts aren't real ghosts! They aren't supposed to be polled for midround roles, and doing so prevents them from returning to their bodies.


Agateophobia (approved):
This should work, there's no reason it shouldn't. If it doesn't work, I will kill myself.

PR #523, opened 2025-01-13
Bites with bloody lover now looks like a kiss

JohnChiffir:
## About The Pull Request
I WAS MADE FOR LOVING YOU BABY

With "Bloody lover" quirk, you're "kissing" instead of biting. It should help for bystanders to understand that in fact, you just love a random hobbo and you're not drinking his blood.
image

## Changelog
changed text and font appearing while biting with bloody lover quirk... I mean I wrote it like three times already


Metekillot:
So I looked at the code and I see that the former maintainers copy pasted it so we're going to need to get rid of that but what you've done here is fine. I like it.

Metekillot (commented):
I approve this but I need to clean up that code a little bit

Metekillot (approved):
Messaging is good. I'll be fixing the code as is anyway...

PR #534, opened 2025-01-15
Stopgap solution to Garou not getting ATM PINs

faaaay:
## About The Pull Request

I put up #497 to update the About Me button for Garou to current standards + allow them to access their own bank accounts, but it was made shortly before the species code freeze and can't be merged. This is a short-term solution that causes their cards to disappear and places cash in their hand instead.

## Why It's Good For The Game

Garou currently effectively spawn with 0 dollars because they can't use their own bank account. This fixes that, and also makes IC sense for some werewolves who wouldn't necessarily have a bank account!

## Changelog
:cl:
tweak: garou now spawn with cash instead of a credit card that they can't use
/:cl:

### DIFF
--- a/code/modules/vtmb/electronics/atms/atm.dm
+++ b/code/modules/vtmb/electronics/atms/atm.dm
@@ -88,6 +88,11 @@ var/mob/living/carbon/human/H
 
 /obj/item/vamp/creditcard/New(mob/user)
 	..()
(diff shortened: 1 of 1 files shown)


Metekillot:
We're not going to do any more stopgap fixes because the code base is a pile of stopgap fixes

PR #536, opened 2025-01-17
Fireman carry quick fix

cheapcents:
## About The Pull Request

The proc originally used to move mobs buckled to another for some reason broke grabs and unbuckled the mob from anything they were on, making fireman carries completely useless. I replaced it for the proc it would usually call right after unbuckling the mob

**However, this for some reason disables diagonal movement while carrying someone.** I didn't bother trying to find a way to safely use the proper movement functions

## Why It's Good For The Game

https://github.com/user-attachments/assets/8c25097d-ce45-43f9-b062-6e4e57fc89b5

## Changelog
:cl:
add: Fireman carry is fixed
:cl:

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -769,7 +769,7 @@
 	for(var/m in buckled_mobs)
 		var/mob/living/buckled_mob = m
 		if(!buckled_mob.Move(newloc, direct, glide_size_override))
-			forceMove(buckled_mob.loc)
+			doMove(buckled_mob.loc) //forcemove has an override that breaks grabs and unbuckles, for some reason
 			last_move = buckled_mob.last_move
 			inertia_dir = last_move
 			buckled_mob.inertia_dir = last_move


Metekillot:
I haven't tested but are you saying this breaks diagonal movement if you're performing a fireman carry?

cheapcents:
diagonal movement is completely disabled while you're carrying someone, but it's fixed once you drop the person being carried

Metekillot:
So we can't fix something if it breaks something else we need to figure out how to just fix the problem without introducing a new one. 

We're not doing any more stopgaps or temp fixes

cheapcents:
Understood, will mark this as draft and try to find a more permanent fix tomorrow

cheapcents:
@Metekillot the reason piggybacking and fireman carrying do not work properly is actually part of a much bigger problem with the move proc for all movable atoms and its return value

im not doing this today

Metekillot:
> @Metekillot the reason piggybacking and fireman carrying do not work properly is actually part of a much bigger problem with the move proc for all movable atoms and its return value
> 
> im not doing this today

Rome wasn't built in a day

cheapcents:
closing this since porting the movement fix from tgstation would genuinely be more practical

PR #545, opened 2025-01-19
Cleans up obfuscate hotfix to be more reasonable about what it reveals, adds several signals

Metekillot:
## About The Pull Request

My obfuscate hotfix is over-reacting to any clicks at all. I added some signals to mitigate that, and these signals are also agnostic, so they can be used in other instances.

## Why It's Good For The Game

Fixing the hotfix.

## Changelog
:cl: Bisar
fix: Obfuscate will be less sensitive about what will reveal you
/:cl:


Nyvrem:
Will this also fix things like smashing in windows/smashing up cars/pouring out gasoline not breaking obfuscate?

wgtjunior743:
Also it is been reported that people can still obfuscate and use guns and it won't trigger it being disabled, not sure if this will fix it.

PR #549, opened 2025-01-19
Fixes say without breaking hotkey say, fixes coloring on input bar

Metekillot:
## About The Pull Request

As the title says; you can type in the bottom input bar if you prefer to be able to press tab to enter communication instead of using the hotkey.

## Changelog
:cl: Bisar
fix: fixes coloring and an input command
/:cl:

### DIFF
--- a/code/__DEFINES/colors.dm
+++ b/code/__DEFINES/colors.dm
@@ -1,7 +1,7 @@
 // This is eventually for wjohn to add more color standardization stuff like I keep asking him >:(
 
-#define COLOR_INPUT_DISABLED			"#F0F0F0"
-#define COLOR_INPUT_ENABLED				"#D3B5B5"
+#define COLOR_INPUT_DISABLED			"#000000"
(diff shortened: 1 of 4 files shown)


cheapcents:
I tried to do this before, but it caused clients with a preexisting save file to lose the keybind. Does this happen here?

Metekillot:
> I tried to do this before, but it caused clients with a preexisting save file to lose the keybind. Does this happen here?

Do you have the branch still up? We'd be able to compare code if so; it worked for me on my testing and as I understand it, testing deployments work off of save files just the same.

cheapcents:
> Do you have the branch still up? We'd be able to compare code if so; it worked for me on my testing and as I understand it, testing deployments work off of save files just the same.

ive since deleted the branch, but my implementation involved directly editing the verb called by the hotkey whereas you simply changed the client proc

if you didn't immediately get a warning upon starting the game that an update caused one of your keybinds to be unbound, then this is probably good to merge without issue

Miraviel (approved):

PR #554, opened 2025-01-20
enables shift-rightclick creating the dropdown menu

Metekillot:
## About The Pull Request

This menu is very useful for many reasons, not least of all for admin purposes, as such, I think it should be -

## Why It's Good For The Game
Lessens our suffering in this life if only for a moment.

## Changelog
:cl: Bisar
add: Shift rightclick will create the popup menu again.
/:cl:


Agateophobia (approved):
They said beauty is dead in the world. I say approved.

PR #567, opened 2025-01-22
Disable Atmos, Part 1

TheCarnalest:
## About The Pull Request
Stops the Atmospherics and Atmospherics Adjacency subsystems from firing while setting all turfs as planetary_turf, essentially implementing static atmos and removing all post-roundstart atmospherics lag. This PR is part 1 of completely ripping atmos out of the server, since it's completely unnecessary for our Earth-based maps.

## Why It's Good For The Game
Atmospherics is laggy and I hate it. This gives the server some sweet miliseconds and paves the way for atmospherics' complete removal.

## Changelog
:cl:
code: All turfs are now planetary turfs, meaning atmos is now static.
code: Atmospherics subsystems will no longer fire.
/:cl:


cheapcents:
you're a hero

Metekillot (changes requested):
Are you alright with not modeling, say, the heat from the fire of a burning building? We'll need to re-implement that ourselves if we go ahead with disabling the atmos system. It won't hurt, of course, because you are right that we don't really need it. But I do want to stress that it's a common misconception that the most expensive system in /tg/ core is atmos -- usually, it's things like Life() which in our case, is eminently true. As well, we have a bunch of snowflake mob movement subsystems that really didn't need to be their own subsystems but that's beside the point of your PR :)

I tested it out; Humans still breathe, so tiles will slowly get filled with CO2 if they're not specifically set to remain completely and totally static. Humans also still suffocate in areas with no oxygen, but maybe we can use that for underwater shit?

Let me know what your vision is with optimization and why we chose atmos as the first one?

Metekillot (approved):
Good enough for me. We have the Methuselah Atmos anyway and I'd rather we chop off as much dead weight as we can; our interests align! This coalition is ironclad. We shall sweep the streets, collecting the lost and lame into our steadily growing horde

PR #570, opened 2025-01-23
Fix Torpor bug

TheCarnalest:
## About The Pull Request
A bug was introduced in the Kuei-jin PR where, for some reason, Torpor code was copy and pasted with no regard for the fact that it was an if/else. This broke torpor for Kindred while letting it work for Kuei-jin. This PR removes the copy and pasted code to let it work for both.

## Why It's Good For The Game
The bug is annoying and disables a fun feature.

## Changelog

:cl:
fix: Torpor works for Kindred again.
/:cl:


cheapcents (approved):
Compiles and runs without issue. Got myself killed as a vampire and successfully revived through torpor

PR #574, opened 2025-01-23
Expands what splats dealer can be

AllCorruptingWyrm:
## About The Pull Request

After a conversation about a lack of roles for KJ, I gave it some thought and looked at dealer and thought it would be interesting to let Garou be a wise-guy glasswalker and run the warehouse, or let a black spiral try to start a drug ring within the warehouse too.  So this PR effectively makes the warehouse a wild card and its own power through economic might which seems to be the way of things moving into the future.

## Why It's Good For The Game

Round diversity is, mostly, the number one goal here. The fact that the warehouse isnt guaranteed to be Camarilla aligned will allow more inter-round diplomacy and even the allowance of gimmicks on an easier scale, like a glasswalker dealer of the wise-guy camp.

## Changelog

:cl:
Add: Allows werewolves & wk to be a Dealer.
Changes: The job text when you join.
/:cl:

### DIFF
--- a/code/controllers/subsystem/job.dm
+++ b/code/controllers/subsystem/job.dm
@@ -568,7 +568,7 @@ SUBSYSTEM_DEF(job)
 
 		to_chat(M, "<b>As the [rank] you answer directly to [job.supervisors]. Special circumstances may change this.</b>")
 		var/mob/living/carbon/human/human = living_mob
-		if((iskindred(human) && human.clane) || iscathayan(human) || isgarou(human))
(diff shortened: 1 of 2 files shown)


cheapcents (changes requested):
Hey there,

v_duty will only log if the player is a kindred. implement alternate properties for logging to garou or kuei-jin without affecting the other jobs and i'll get this merged

you can see this functionality @ code\controllers\subsystem\job.dm, line 571

AllCorruptingWyrm:
Were good for a merge, I think after I was shown a runtime.

cheapcents (approved):

PR #581, opened 2025-01-24
Fix NPCs killing each other

TheCarnalest:
## About The Pull Request
Fixes an oversight in the faction removal PR that allowed NPCs to aggro on each other.

## Why It's Good For The Game
Cops killing each other is funny, but not very good.

## Changelog

:cl:
fix: NPCs no longer kill each other.
/:cl:


Metekillot (approved):
i love for police to be the ones experiencing police violence for once but it just is not immersive. we need to think of the immersion.

PR #585, opened 2025-01-24
Potence knockback now happens on disarm instead of attack

boy2mantwicethefam:
Because knocking away the people you want to beat up is pretty bad.

### DIFF
--- a/code/modules/mob/living/carbon/carbon_defense.dm
+++ b/code/modules/mob/living/carbon/carbon_defense.dm
@@ -386,6 +386,9 @@
 			else
 				append_message = "loosening [target.p_their()] grip on [target_held_item]"
 		log_combat(src, target, "shoved", append_message)
+	if(potential >= 5)
+		var/atom/throw_target = get_edge_target_turf(target, shove_dir)
+		target.throw_at(throw_target, rand(5, 7), 4, src, gentle = TRUE) //No stun nor impact damage from throwing people around
 
 /mob/living/carbon/proc/is_shove_knockdown_blocked() //If you want to add more things that block shove knockdown, extend this
 	for (var/obj/item/clothing/clothing in get_equipped_items())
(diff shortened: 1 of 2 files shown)


Metekillot:
We'll need to hold a maintainer discussion for this first

Metekillot:
The knock back serves as an indication of what is a masquerade violating display of superhuman strength as well as giving people the brief and often worthless opportunity to escape an instant follow up 60ish damage punch after the one that knocked them back in the first place.

boy2mantwicethefam:
The issue is that the lower levels of Potence don't do the knockback, so people are effectively punished for toggling 5-dot Potence instead of 4-dot Potence despite the relatively marginal increase in power (Potence 4 provides +32 unarmed damage and 2.6x melee damage, Potence 5 provides +40 unarmed damage and 3x melee damage). And since Potence was significantly nerfed they should at least get this in return.

Metekillot:
> The issue is that the lower levels of Potence don't do the knockback, so people are effectively punished for toggling 5-dot Potence instead of 4-dot Potence despite the relatively marginal increase in power (Potence 4 provides +32 unarmed damage and 2.6x item melee damage, Potence 5 provides +40 unarmed damage and 3x item melee damage). And since Potence was significantly nerfed through the removal of the stunlock and wall-impacting they should at least get this in return.

This is a good point; the damage scaling linearly and the knock back being a stark threshold is janky now that you illustrate it like this. I think reworking the calculation to use an entirely different growth function and having knockback be a signal fired by applying damage would actually be a more elegant solution.

boy2mantwicethefam:
@Metekillot I considered it but for now we should probably try knockback-less Potence and see how Potence combat fares with it, plus the noted interest of a triat/project lead in not having knockback happen per punch. If it still feels too overwhelming (such as through Potence + Celerity) we can always try a scaling knockback. Such scaling knockback would also affect the Flesh Shintai discipline which sets the "potential" variable to 4 unless we distinguish Potence potential and Flesh Shintai potential.
@crababot As is there is not much of a reason to activate Potence 5, since it just knocks people away when a Potence user wants to be next to them.

Metekillot:
The linear calculation and arbitrary knockback threshold both need to be comprehensively addressed; pendulous nerfs and buffs that remove this or crank that aren't going to be sufficient to address the issue. I'll likely be closing this with the assurance that as soon as Lucia finishes up the first pull request for the disciplines rework I'll pause what I'm doing to come back and address potence. I'll leave it for another day, however, for people to gather and express their thoughts on it.

Metekillot:
Post-discipline rework we'll revisit potence and give it a growth function to determine knockback instead of it having be a threshold at level 5.

PR #593, opened 2025-01-24
Fixes switchblade having inconsistent stats when created

boy2mantwicethefam:
It had some unintended behavior in that regard, this PR fixes it. It now properly spawns sheathed.


cheapcents (approved):
Good work

PR #594, opened 2025-01-24
Doubles passive experience gain from existing

boy2mantwicethefam:
Issue: The lack of both ambitions and faction wars has led to significantly less exp gained by players throughout the round, forcing them to play certain roles to compensate for it.
Solution: Players will now gain 2 experience points every 45 minutes instead of 1 for existing, which adds up to about as much as either having faction turf dominance for the entire round OR finishing an ambition (3 experience points).

This change should be a decent solution even if there are plans to remove the experience system.


AllCorruptingWyrm:
I beg. Thank you so much

Metekillot:
I'm sorry, and I know it's a colossal pain in the ass. I also wouldn't fault anyone for feeling like it's something of a "fuck you" to newer players floundering against older sweats who have gamed down to gen 7 and (stat > 5), but we can't backtrack on the steps we're taking to remove experience. We, unfortunately, currently lack the developer resources to be able to implement the removal all at once so there will be some growing pains as we work toward doing so.

PR #599, opened 2025-01-25
Protean Walancing: Try One

SELFHELL:
## About The Pull Request

Rebalances Protean,
Protean Claws go from applying **30** Clone Damage on hit, To applying **15** Clone Damage and **15** Brute Damage
Rebalances the Protean Warform Healthpools

## Why It's Good For The Game
Lets talk some numbers here, 

The baseline Garou attack in their Crinos form deals a total of **2** Clone Damage

Meanwhile, the first level of Protean provides users with `/obj/item/melee/vampirearms/knife/gangrel`, a humble weapon dealing a total of 30 points almost unhealable clone damage with **100%** AP.

I don't think any Protean 1 user should be dealing Fifteen times the clone damage done by garou. This PR gives them a far more reasonable number of 15 CL, with 15 much easier to heal brute. If the value is STILL too high, I'll adjust as needed.

As for the health changes, the war-form health really shouldn't be scaling that rapidly.

## Changelog

:cl:
balance: Rebalances Protean
/:cl:

### DIFF
--- a/code/modules/vtmb/vampire_clane/tzimisce.dm
+++ b/code/modules/vtmb/vampire_clane/tzimisce.dm
@@ -709,8 +709,8 @@
 	mob_size = MOB_SIZE_HUGE
 	speak_chance = 0
 	speed = -0.4
-	maxHealth = 400
-	health = 400
+	maxHealth = 350
+	health = 350
 	butcher_results = list(/obj/item/stack/human_flesh = 10)
 	harm_intent_damage = 5
 	melee_damage_lower = 40
@@ -728,8 +728,8 @@
 	possible_a_intents = list(INTENT_HELP, INTENT_GRAB, INTENT_DISARM, INTENT_HARM)
 
 /mob/living/simple_animal/hostile/gangrel/better
-	maxHealth = 500
-	health = 500
+	maxHealth = 400
+	health = 400
 	melee_damage_lower = 45
 	melee_damage_upper = 45
 	speed = -0.6
@@ -737,8 +737,8 @@
 /mob/living/simple_animal/hostile/gangrel/best
 	icon_state = "gangrel_m"
 	icon_living = "gangrel_m"
-	maxHealth = 600
-	health = 600
+	maxHealth = 500
+	health = 500
 	melee_damage_lower = 55
 	melee_damage_upper = 55
 	speed = -0.8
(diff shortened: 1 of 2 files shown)


JohnChiffir:
The discipline itself isn't good and the only part that still holds it is those 30 clone dmg, by lowering it to 15 it's MUCH more convenient to just use potence 5 + any melee weapon (it'll be much more efficient and I ain't even talking bout celerity combo). 

And HP for warforms... someone doesn't even know that all damage done to the warform tanslates to the carbon after the end of discipline, so effective HP for all warforms is just 350 (7 gen + 5 physique)... so why this change was even made?

If you want to change numbers (as we can't do anything significant before the rework). You can make for protean 1 lvl to make 15 clone dmg and for protean 2 — ~25, as for now the only difference between them is only +10% movement speed  

But who the hell I am to say that...

MercuryArrow:
While the 30 clone can be *rough* to deal with in terms of attrition, that's kind of the point - aggravated damage weapons are meant to wear you down over time and be more of a hit and run weapon, and they do that job just fine. Even in their current form, they're still outclassed by many, MANY of our higher-end weapon types which can burst even Kindred down more effectively.

The historical sentiment has been that Gangrel have always had it pretty bad in terms of disciplines(Protean is effectively 2 powers instead of 5, Animalism has varied a lot over time, Fortitude is decent but incredibly-short-acting and hard to get feedback off of). Couple that with the fact their claws are _the worst clone-damage-focused ability in the codebase_ despite that traditionally being their main combat niche in the tabletop and this won't feel satisfying for anyone.

JohnChiffir:
The other way to balance things out, is to lock protean to gangrels only. In that case ammount of protean users will be vastly reduced (cuz for now we have: protean users > gangrels), making encounters against it more rare in the same way as we have for WL clans (which have OP discs as well)

Metekillot:
We're close enough to the release of the refactor in order for us to want to hold off on fucking with the numbers on any combat shit because elder vampires are soon going to have much different levels of HP

SELFHELL:
> We're close enough to the release of the refactor in order for us to want to hold off on fucking with the numbers on any combat shit because elder vampires are soon going to have much different levels of HP

Sounds good, i'll hold off on this then

PR #607, opened 2025-01-26
Changes sniper rifle caliber to .50, massively increases damage

boy2mantwicethefam:
It's not exactly a new unique caliber but that's because I thought I'd have to add this caliber to the game.
Changed the sniper rifle's ammo to .50 bullets, which are kinda meant to be used with sniper rifles anyway.
Changed the damage multiplier from 1.5 to 2.
In total this makes the sniper rifle go from 67.5 damage to 140 damage. Welcome back crossbows (keep in mind they have a 4 second fire delay between shots without Gunslinger, while Crossbows could fire immediately after).

To-do: Replace the associated 5.56 ammo boxes with .50 ammo boxes wherever a sniper rifle appears.
To-do: Actually test this.

### DIFF
--- a/code/modules/wod13/guns.dm
+++ b/code/modules/wod13/guns.dm
@@ -491,8 +491,8 @@
 /obj/item/ammo_box/magazine/internal/vampire/sniper
 	name = "sniper rifle internal magazine"
 	desc = "Oh god, this shouldn't be here"
-	ammo_type = /obj/item/ammo_casing/vampire/c556mm
-	caliber = CALIBER_556
+	ammo_type = /obj/item/ammo_casing/vampire/c50
+	caliber = CALIBER_50
 	max_ammo = 5
 	multiload = TRUE
 
@@ -522,7 +522,7 @@
 	zoom_amt = 10 //Long range, enough to see in front of you, but no tiles behind you.
 	zoom_out_amt = 5
 	slot_flags = ITEM_SLOT_BACK
-	projectile_damage_multiplier = 1.5
+	projectile_damage_multiplier = 2 //140 damage. Nice.
 	actions_types = list()
 	masquerade_violating = TRUE
 	cost = 250


Metekillot:
Replace the 5.56 where necessary and I say this is good to go. I'll PR replacing all the 5.56 mags and boxes and then I'll say this is good to go.

Metekillot:
Also it looks like flav never actually implemented sprites for .50 so we'll need to fix that as well

Metekillot (changes requested):
see comments

Metekillot:
I've made the in hand sprites for the bullets and the box, I'll PR those shortly then I would appreciate if you did the rest of the sprites; the remainder for each is like, 2-6 pixels per

code/modules/wod13/ammostack.dm
```diff
@@ -276,6 +276,12 @@
 	ammo_type = /obj/item/ammo_casing/vampire/c44
 	max_ammo = 60
 
+/obj/item/ammo_box/vampire/c50
+	name = "ammo box (.50)"
+	icon_state = "50box"
+	ammo_type = /obj/item/ammo_casing/vampire/c50
+	max_ammo = 30
```
cheapcents, line 283: 30 cartridges of insta-critting beowulf for just 1.5k sounds a bit excessive...

let's lower this to 20 bullets and increase the price to around 2.5k, whatever you think is best

code/modules/wod13/guns.dm
```diff
@@ -522,7 +522,7 @@
 	zoom_amt = 10 //Long range, enough to see in front of you, but no tiles behind you.
 	zoom_out_amt = 5
 	slot_flags = ITEM_SLOT_BACK
-	projectile_damage_multiplier = 1.5
+	projectile_damage_multiplier = 2 //140 damage. Nice.
```
cheapcents, line 525: gaming

cheapcents (changes requested):

XanderDox:
@boy2mantwicethefam are you still working on this?

Miraviel (changes requested):
code looks good, it will need a map rebase after #679 is merged

Metekillot:
Merged in #815

PR #612, opened 2025-01-26
Reduces drug lab cost by 60%

boy2mantwicethefam:
## About The Pull Request
Reduced the drug lab's cost from $10000 to $4000.

## Why It's Good For The Game
The drug lab is extremely expensive and takes way too much luck to make up for its cost compared to something like selling green packages. It also locks anyone who doesn't have at least Social 5 in it away from making any sort of long-term profit. Very unprofitable, mainly due to its steep initial cost, which this PR reduces.
## Changelog

:cl:
balance: The drug lab's price is now $4000 instead of $10000.
/:cl:

### DIFF
--- a/code/modules/wod13/cargo.dm
+++ b/code/modules/wod13/cargo.dm
@@ -21,7 +21,7 @@
 /datum/supply_pack/vampire/methlab
 	name = "Lab Equipment"
 	desc = "Contains lab equipment."
-	cost = 10000
+	cost = 2000
 	contains = list(/obj/structure/methlab/movable)
 
 /datum/supply_pack/vampire/gas_can
--- a/code/modules/wod13/cargo.dm
+++ b/code/modules/wod13/cargo.dm
@@ -21,7 +21,7 @@
 /datum/supply_pack/vampire/methlab
 	name = "Lab Equipment"
 	desc = "Contains lab equipment."
-	cost = 2000
+	cost = 4000
 	contains = list(/obj/structure/methlab/movable)
 
 /datum/supply_pack/vampire/gas_can


AllCorruptingWyrm:
Yeah this is about the same math I had on my PR.
I'm not sure how I feel about more money entering the economy but eeh..

code/modules/wod13/cargo.dm
```diff
@@ -21,7 +21,7 @@
 /datum/supply_pack/vampire/methlab
 	name = "Lab Equipment"
 	desc = "Contains lab equipment."
-	cost = 10000
+	cost = 2000
```
Metekillot, line 24:
```suggestion
	cost = 4000
```
12-13 to ~5 seems more reasonable to begin receiving a return on investment.

Metekillot, line 24:
```suggestion
	cost = 4000
```

Metekillot (changes requested):
12-13 to ~5 seems more reasonable to begin receiving a return on investment.

code/modules/wod13/cargo.dm
boy2mantwicethefam: The bag numbers were done with a maxed out Social

XanderDox: @Metekillot thoughts? ^

XanderDox:
Please add a proper changelog for the updates channel.

code/modules/wod13/cargo.dm
Metekillot: > The bag numbers were done with a maxed out Social, at 1 Social the payout is basically 1/5th. It would make blue bag creation outright unprofitable at 1 and maybe 2 Social.

Having low social skills meaning you don't make money from things that require social skills means you need to rely on another player with social skills, which I'm fine with.

boy2mantwicethefam: Hmm, sure let's do it this way.

Miraviel (approved):

PR #621, opened 2025-01-26
Improves Admin Global Narrate

silencer-pl:
## About The Pull Request

This PR changes the admin "Global Narrate" command in the following ways:

- The command now asks if the user wishes to use a header, then asks them to enter the header (if selected) and body of the message.
- The command instructs the user about HTML tags and paragraphs where relevant
- The command prints out a test display of the resulting message and asks the user for final verification.
- The command broadcasts the entered sequence to the world.

<details><summary>Click for Example</summary>
<p>

image

</p>
</details>

Two potential tweaks:

- Removal of the narration question dialog and leaving that just to the header_input box. This however seems very sloppy, to me at least.
- Removal of the preview option. Though maybe I can pull that into a client toggle later if needed.

Both of these likely can be made as admin setting toggles or something, but that's going a tad out of scope than I intended, it is an option though.

## Why It's Good For The Game

There was an admin roleplaying as a DJ and using the in game music player to play songs yesterday and I felt like their messages could very much get lost if someone was not already paying attention. This adds some punch up to the text and generally makes it more readable (walls of text tend to look better justified, which is why the body is that) and also lets admins add an additional, mood setting header which is optional. 

Better mood/narration controls means better admin narrations, means better RP means happier players ;P

## Changelog

- Made admin narration more pronounced and offer an optional header.

### DIFF
--- a/code/__DEFINES/chat.dm
+++ b/code/__DEFINES/chat.dm
@@ -18,3 +18,7 @@
 #define MESSAGE_TYPE_ADMINLOG "adminlog"
 #define MESSAGE_TYPE_ATTACKLOG "attacklog"
 #define MESSAGE_TYPE_DEBUG "debug"
+
+// Global Narrate
+#define narrate_head(str) ("<div class='narrate_head'><p>" + str + "</p></div>")
+#define narrate_body(str) ("<div class='narrate_body'><p>" + str + "</p></div>")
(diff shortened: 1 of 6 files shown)


silencer-pl:
After getting some answers about the future of this codebase, I am no longer interested in contributing code and am withdrawing this MR, thank you and sorry for the trouble.

PR #628, opened 2025-01-26
Bear and Bear Cave. 

CactusMouth:
## About The Pull Request
image
image

'I do not have to outrun the bear, Ted. I just have to outrun you.'

Adds a Bear (Sprites found in 64x64, did not make it myself) simplemob, quite strong requiring several people or one skilled player to defeat. May cop players weep. Also adds a Bear Cave in replacement for a random homeless cave. Just South of the tunnel from the Manor.

## Why It's Good For The Game

Never leave a sprite unused. Gives people a chance to test out their skill and for admins to play with. I would've added it as an Animalism module but I don't want to touch that code and do not want to make it overpowered. As a random simplemob, it can be as powerful as reasonably possible without disrupting balance.

## Changelog

:cl:
add: Bear
map: Replace the random homeless area with a new, fun bear cave.
code: Simplemob 'Bear'. New file
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/werewolf/transformation.dm
+++ b/code/modules/mob/living/carbon/werewolf/transformation.dm
@@ -52,7 +52,7 @@
 			H.remove_overlay(PROTEAN_LAYER)
 			G.punchdamagelow = G.punchdamagelow-15
 			G.punchdamagehigh = G.punchdamagehigh-15
-			H.physique = initial(H.physique)
+			H.physique = H.physique-2
 			H.physiology.armor.melee = H.physiology.armor.melee-15
 			H.physiology.armor.bullet = H.physiology.armor.bullet-15
 			var/matrix/M = matrix()
--- a/code/modules/vtmb/werewolf/gifts.dm
+++ b/code/modules/vtmb/werewolf/gifts.dm
@@ -448,7 +448,7 @@
 			H.remove_overlay(PROTEAN_LAYER)
 			G.punchdamagelow -= 15
 			G.punchdamagehigh -= 15
-			H.physique = initial(H.physique)
+			H.physique = H.physique-2
 			H.physiology.armor.melee -= 15
 			H.physiology.armor.bullet -= 15
 			var/matrix/M = matrix()


UltraBigRat:
What is this bear on, guy has over double the damage and over triple the health of a Crinos Warform for some reason.

Eluxor:
I would make the cave deeper and have some sort of door or obstacle. Otherwise, the bear will always come out. Mobs randomly walk and he might get out and maul a random passerby.

CactusMouth:
Does the NPC block not work on Simple Mobs?

CactusMouth:
Done, updated it a bit

code/modules/wod13/bear.dm
```diff
@@ -0,0 +1,35 @@
...
+	response_help_continuous = "pokes"
+	response_help_simple = "poke"
+	response_disarm_continuous = "gently pushes"
+	response_disarm_simple = "gently push"
+	response_harm_continuous = "punches"
+	response_harm_simple = "punch"
+	can_be_held = FALSE
+	density = FALSE
+	anchored = FALSE
+	footstep_type = FOOTSTEP_MOB_CLAW
```
Miraviel, line 24:
```suggestion
```

code/modules/wod13/bear.dm
```diff
@@ -0,0 +1,35 @@
...
+
+	bloodquality = BLOOD_QUALITY_LOW
```
Miraviel, line 26:
```suggestion
```

code/modules/wod13/bear.dm
```diff
@@ -0,0 +1,35 @@
...
+	bloodpool = 1
+	maxbloodpool = 1
+	del_on_death = 1
```
Miraviel, line 29:
```suggestion
```

If you have this, there is no corpse - when they die, they get deleted. If you want this to happen for some reason, you don't need a dead sprite, nor the `butcher_results` above because the mob will never be butcherable!

code/modules/wod13/bear.dm
```diff
@@ -0,0 +1,35 @@
...
+	response_harm_continuous = "punches"
+	response_harm_simple = "punch"
+	can_be_held = FALSE
+	density = FALSE
+	anchored = FALSE
+	footstep_type = FOOTSTEP_MOB_CLAW
+
+	bloodquality = BLOOD_QUALITY_LOW
+	bloodpool = 1
+	maxbloodpool = 1
```
Miraviel, line 28:
```suggestion
```

Probably not a good idea to override these. The default mob bloodpool is 5, rats have 1. This would make bears have as much blood as rats!

code/modules/wod13/bear.dm
```diff
@@ -0,0 +1,35 @@
...
+	del_on_death = 1
+	maxHealth = 850
+	health = 850
+	cached_multiplicative_slowdown = 2
```
Miraviel, line 32:
```suggestion
```

This variable shouldn't be declared on init, it is calculated after spawning. If you want to edit the mob's speed, you need to edit its `speed` (among other things, but bears already have a speed of 0).

code/modules/wod13/bear.dm
```diff
@@ -0,0 +1,35 @@
...
+	density = FALSE
+	anchored = FALSE
+	footstep_type = FOOTSTEP_MOB_CLAW
+
+	bloodquality = BLOOD_QUALITY_LOW
+	bloodpool = 1
+	maxbloodpool = 1
+	del_on_death = 1
+	maxHealth = 850
+	health = 850
```
Miraviel, line 31:
```suggestion
	maxHealth = 150
	health = 150
```

You might want to reconsider these values. Our current Beefy Boys:tm:, the Crinos form of werewolves have a health pool of 250. Afaik Potence currently does not work on simplemobs (like this one), nor does clone damage with which werewolves are buffed with, so this bear would mostly be killable with guns only.

The suggested values are just a spitball. Spawn in as a few characters who would likely encounter this creature and try to fight it, see what feels challenging, but still doable?

code/modules/wod13/bear.dm
```diff
@@ -0,0 +1,35 @@
...
+	cached_multiplicative_slowdown = 2
+
+	melee_damage_lower = 50
+	melee_damage_upper = 60 //Good luck lol
```
Miraviel, line 35:
```suggestion
	melee_damage_lower = 35
	melee_damage_upper = 40 //Good luck lol
```

These values would take out a garou in their warform in less than ten seconds. See the comment above at the health regarding this!

You can also check for other mobs and see how much they damage. The values I suggested are equal to the level 5 Giovanni zombies, but it is better to test it ingame if this whoops people's asses too fast or not.

code/modules/wod13/bear.dm
CactusMouth: Minor mistake here on my end, thought I set it to 0.

code/modules/wod13/bear.dm
CactusMouth: I'm going to try to test some of this stuff out to get a good sense on it's health. This bear should be a sort of mini boss-battle for the loot inside it's cave (Shotgun, money and vamp artifact) but still beatable with teamwork/experience.

CactusMouth:
Going to do some testing, then send out a new reform for the bear based on it's health.

CactusMouth:
Done some editing, I've found around 500hp is ideal for what I'm looking for in the mob. This should be somewhat difficult to most people that aren't heavily armed or skilled, and would be ideal for around 2-3 people to fight. Damage has been decreased however.

Ghastlytrench:
Bear Mini-boss.

CactusMouth:
More or less. I wanted something that could actually pose a danger to people but sadly I'm not a good enough coder to add anything substantial.

CactusMouth:
Removing the bear cave from the standard PR due to the griefing potential and simple mob AI, I'll leave it at just a bear that can be spawned by an Admin.

PR #630, opened 2025-01-27
Cruelty will now respect gun damage multipliers

boy2mantwicethefam:
Turns out it would reset the damage back to the initial value multiplied by 2, which messes with 3 different guns that have damage multipliers (snub-nosed revolvers, sniper rifles and the semi-auto shotguns). This fixes it.

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -576,7 +576,7 @@
 			var/mob/living/carbon/human/frer = firer
 			if(frer.blood)
 				if(prob(frer.get_total_blood()*10))
-					damage = initial(damage)*2
+					damage *= 2
 	if(temporary_unstoppable_movement)
 		temporary_unstoppable_movement = FALSE
 		movement_type &= ~PHASING


Metekillot:
Cruelty is already relatively overpowered as is and adding an additional *2 damage multiplier to any possible gun damage multiplier is obscene.

code/modules/projectiles/projectile.dm
```diff
@@ -576,7 +576,7 @@
 			var/mob/living/carbon/human/frer = firer
 			if(frer.blood)
 				if(prob(frer.get_total_blood()*10))
-					damage = initial(damage)*2
+					damage *= 2
```
Metekillot, line 579:
```suggestion
					damage == initial(damage) ? damage * 2 : clamp(damage, initial(damage), initial(damage) * 2)
```
Cruelty should give you double damage of a projectile, at max; not double the damage of any given projectile, including ones that are possibly already increased. Even still, we may nerf it later, because a % double damage chance per stat point is ludicrous, but we'll worry about that down the line.

Metekillot (changes requested):

boy2mantwicethefam:
Cruelty is comparatively expensive XP-wise though. This is pretty much a bugfix because the whole thing is giving me the impression that it was a way to double-check just in case it somehow got past the variable check, which never actually happens in practice as far as I know. Unless someone from VTM13 says "oh yeah we did this so that damage multipliers don't get too crazy" then I think it's a bugfix. On top of that leaving it as is is not that great for futureproofing either; if a gun with a very low damage multiplier gets added it will end up with a way higher damage multiplier from Cruelty than it should, and inversely a gun with a damage multiplier greater than 2 will turn Cruelty into a critical fail chance and deal reduced damage. Would rather leave the PR as is.

Metekillot:
We don't care what the maintainers of VtM13 intended.

boy2mantwicethefam:
Technically, but the rest of the post still stands. Better to let Cruelty actually work out as it should rather than introduce unusual limitations to it, the guns themselves can always be nerfed if they get excessive.

SoreYew:
Considering the Sniper is a laughable joke and the snub nose is 9mm, if Cruelty worked as intended then these weapons would have actual use. 

Are the semi shotties even implemented?

Metekillot:
> Considering the Sniper is a laughable joke and the snub nose is 9mm, if Cruelty worked as intended then these weapons would have actual use.
> 
> Are the semi shotties even implemented?

The sniper will be receiving a significant and sorely needed buff. If you're referring to the SPAS; currently it's exclusive to police requisitions but we'll be adjusting firearms availability as we move forward.

Metekillot:
what the fuck when did i merge this

Metekillot:
what the fuck???

Metekillot:
Eh we're refactoring this shit anyway

PR #631, opened 2025-01-27
Gunfighter now only reduces delay by half instead of removing it

boy2mantwicethefam:
## About The Pull Request
Changes the Gunfighter archetype's delay reduction from "basically no delay between shots" to "halved delay".
Slightly reworks the code so that it uses a trait instead of a variable.

## Why It's Good For The Game
It's somewhat more balanced. Rapid fire guns will still fire very fast but they should no longer literally fire like shotguns. Also makes sure the [buffed sniper rifle](https://github.com/WorldOfDarknessXIII/World-of-Darkness-13/pull/607) does not become OP (its main drawback is a 4 second delay between shots which gets removed with Gunfighter).

## Changelog
:cl:
balance: The Gunfighter archetype now only reduces firing delays by half instead of removing them.
/:cl:

### DIFF
--- a/code/modules/mob/living/living_defines.dm
+++ b/code/modules/mob/living/living_defines.dm
@@ -202,7 +202,7 @@
 	var/last_taste_time
 	var/last_taste_text
 
-	var/no_fire_delay = FALSE
+	var/halved_fire_delay = FALSE
 
 	var/experience_plus = 0
 	var/discipline_time_plus = 0
--- a/code/modules/mob/living/living_defines.dm
+++ b/code/modules/mob/living/living_defines.dm
@@ -202,8 +202,6 @@
 	var/last_taste_time
 	var/last_taste_text
 
-	var/halved_fire_delay = FALSE
-
 	var/experience_plus = 0
 	var/discipline_time_plus = 0
 	var/bloodpower_time_plus = 0
--- a/code/modules/projectiles/ammunition/_firing.dm
+++ b/code/modules/projectiles/ammunition/_firing.dm
@@ -16,14 +16,13 @@
 		AddComponent(/datum/component/pellet_cloud, projectile_type, pellets)
 		SEND_SIGNAL(src, COMSIG_PELLET_CLOUD_INIT, target, user, fired_from, randomspread, spread, zone_override, params, distro)
 
-	var/mob/living/carbon/human/shooter = user
 	if(click_cooldown_override)
-		if((click_cooldown_override > CLICK_CD_RAPID) && HAS_TRAIT(shooter, TRAIT_GUNFIGHTER))
+		if((click_cooldown_override > CLICK_CD_RAPID) && HAS_TRAIT(user, TRAIT_GUNFIGHTER))
 			user.changeNext_move(max(CLICK_CD_RAPID, round(click_cooldown_override/2)))
 		else
 			user.changeNext_move(click_cooldown_override)
 	else
-		if(HAS_TRAIT(shooter, TRAIT_GUNFIGHTER))
+		if(HAS_TRAIT(user, TRAIT_GUNFIGHTER))
 			user.changeNext_move(CLICK_CD_RAPID)
 		else
 			user.changeNext_move(CLICK_CD_RANGE)
--- a/code/modules/projectiles/ammunition/_firing.dm
+++ b/code/modules/projectiles/ammunition/_firing.dm
@@ -17,14 +17,13 @@
 		SEND_SIGNAL(src, COMSIG_PELLET_CLOUD_INIT, target, user, fired_from, randomspread, spread, zone_override, params, distro)
 
 	var/mob/living/carbon/human/shooter = user
-	var/less_firing_delay = shooter.halved_fire_delay
 	if(click_cooldown_override)
-		if((click_cooldown_override > CLICK_CD_RAPID) && less_firing_delay)
+		if((click_cooldown_override > CLICK_CD_RAPID) && HAS_TRAIT(shooter, TRAIT_GUNFIGHTER))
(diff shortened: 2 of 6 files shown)


code/modules/mob/living/living_defines.dm
```diff
@@ -202,7 +202,7 @@
 	var/last_taste_time
 	var/last_taste_text
 
-	var/no_fire_delay = FALSE
+	var/halved_fire_delay = FALSE
```
Metekillot, line 205: Just remove this extra variable entirely. We don't need to have a variable on living just to determine fire rate.

code/modules/projectiles/ammunition/_firing.dm
```diff
@@ -18,14 +18,14 @@
 
 	if(click_cooldown_override)
 		if(click_cooldown_override > CLICK_CD_RAPID)
-			if(user.no_fire_delay)
+			if(user.halved_fire_delay)
 				user.changeNext_move(max(CLICK_CD_RAPID, round(click_cooldown_override/2)))
```
Metekillot, line 22: Have the resolution of fire rate be the result of a returned bitfield from a signal. Take a look at how other signals return bitfields for an example.

code/modules/projectiles/ammunition/_firing.dm
```diff
@@ -18,14 +18,14 @@
...
 			else
 				user.changeNext_move(click_cooldown_override)
 		else
 			user.changeNext_move(click_cooldown_override)
 	else
-		if(user.no_fire_delay)
+		if(user.halved_fire_delay)
 			user.changeNext_move(CLICK_CD_RAPID)
```
Metekillot, line 29: Why do we check for this twice?

code/modules/projectiles/gun.dm
```diff
@@ -322,8 +322,8 @@
 	var/randomized_bonus_spread = rand(0, bonus_spread)
 
 	var/real_fire_delay = fire_delay
-	if(user.no_fire_delay)
-		real_fire_delay = 0
+	if(user.halved_fire_delay)
+		real_fire_delay /= 2
```
Metekillot, line 326: This would be the third time we check for this. We only need to check for it once.

code/modules/vtmb/archetypes.dm
```diff
@@ -50,7 +50,7 @@
 	start_blood = 2
 
 /datum/archetype/gunfighter/special_skill(var/mob/living/carbon/human/H)
-	H.no_fire_delay = TRUE
+	H.halved_fire_delay = TRUE
```
Metekillot, line 53: Have the archetype register a signal on the chooser instead, and have that signal registration call a proc when triggered that returns the bitfield that specifies the fire delay.

Metekillot (changes requested):

code/modules/mob/living/living_defines.dm
Metekillot: So that's usually a result of something going on on your own machine where you need to enable the database

boy2mantwicethefam: I'll look into it on my end. I've also given it some thought and I don't think it being a variable is a bad thing. It's straightforward and it's var-editable, not everything has to be micro-optimized to that degree.

code/modules/projectiles/gun.dm
boy2mantwicethefam: This is a different proc

code/modules/projectiles/ammunition/_firing.dm
boy2mantwicethefam: Because I only replaced the variable names. I'll look into reordering the code a little but this appears to be specifically for whether there's a clicking override. Code-wise neither actually gets checked twice because the conditions that lead to both checks are mutually exclusive.

code/modules/projectiles/gun.dm
Metekillot: If you check for something in a previous proc, or have brought in a return from a previous proc, that gives you that information, you don't need to check for it again in another proc.

code/modules/mob/living/living_defines.dm
Metekillot: It's not about micro optimizing, it's about sensible organization. There's no reason for us to have this one specific variable for one specific quirk on the ENTIRE abstract class of a mob.

boy2mantwicethefam: I'm not sure there's anything in the code that allows for anything other than a human to fire a gun (silicons would be a candidate but they aren't supposed to appear) so I'll probably move the variable to the human defines.

code/modules/projectiles/gun.dm
boy2mantwicethefam: Rewriting the proc to allow passing over the argument would be clutter to be fair.

Metekillot:
Are you going to fight me on every single comment I make when I review your
code?

boy2mantwicethefam:
no

XanderDox:
Please add a proper changelog for the updates channel.

code/modules/vtmb/archetypes.dm
```diff
@@ -50,7 +50,7 @@
 	start_blood = 2
 
 /datum/archetype/gunfighter/special_skill(var/mob/living/carbon/human/H)
-	H.no_fire_delay = TRUE
+	ADD_TRAIT(H, TRAIT_GUNFIGHTER, TRAUMA_TRAIT)
```
Miraviel, line 53:
```suggestion
	ADD_TRAIT(H, TRAIT_GUNFIGHTER, ROUNDSTART_TRAIT)
```

Miraviel (changes requested):

Miraviel (approved):

PR #636, opened 2025-01-28
Removes all times CMNoir/greyscale-vision is set for ghost

mrhugo13:
## About The Pull Request
Removes all times ghost vision is set to or animated towards CMNoir greyscale-o-vision.
Keeps the CMNoir color matrix list incase of future use.

Gif of it not greyscaling your screen when you ghost
dreamseeker_2025-01-28_17-03-03
## Why It's Good For The Game
To be admittedly rather brash, the greyscale vision just makes observing worse, it doesn't add any flavor or theme or any other since of aesthetic but rather wears and tears on your eyes as you try to look at the cool stuff thats happening through a faded lens. Best case it's a nuisance and makes any recording of in-game footage from a ghost/observer perspective hard to look at and at worst it just drives people away from staying to observe or atleast be more interested in a game they're not directly participating in.

## Changelog
:cl: Vondiech/Citruses
del: Ghost greyscale vision is no more.
/:cl:

### DIFF
--- a/code/modules/mob/dead/observer/observer.dm
+++ b/code/modules/mob/dead/observer/observer.dm
@@ -72,11 +72,6 @@ var/list/CMNoir = list(0.3,0.3,0.3,0,\
 	var/aghosted = FALSE
 	var/auspex_ghosted = FALSE
 
-/mob/dead/observer/Login()
-	..()
-//There was observer music here. It's gone now.
-//Also ghost greyscaling, straining for the eyes.
-
 /mob/dead/observer/Initialize()
 	set_invisibility(GLOB.observer_default_invisibility)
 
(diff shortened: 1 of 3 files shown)


code/modules/mob/dead/observer/observer.dm
```diff
@@ -74,10 +74,8 @@ var/list/CMNoir = list(0.3,0.3,0.3,0,\
 
 /mob/dead/observer/Login()
 	..()
-	if(client && !aghosted)
-		animate(client, color = CMNoir, time = 30)
-		client.color = CMNoir
```
hex37, line 79: you can remove the whole /mob/dead/observer/Login() wholesale if the only thing it does is call its parent.

mrhugo13: Good point, saving & pushing commit with this bit deleted.

PR #637, opened 2025-01-28
Removed diseases from rotten corpses and organs

JohnChiffir:
## About The Pull Request

Removed diseases from rotten corpses and organs, cuz it doesn't add anything to the gameplay and all diseases require reagents that we don't have

Removed it without commenting as we're getting rid of athmos soon anyway

## Changelog
directly removed miasma sickness


cheapcents (approved):
Good work

PR #640, opened 2025-01-29
Fixes Oversight & Makes more things orderable at the Dealer

AllCorruptingWyrm:
## About The Pull Request
Fixes an oversight from my last PR. The sorting of the vendors caused me to remove the seeds and bailers so now you can get them. Also it sorts the Cargo Vendor from lowest to highest by category & lets magazines be ordered in bulk so you have reason to go to the Dealer too. Silver & Incen cost isn't touched though.

## Why It's Good For The Game

Fixes oversight, less extensive charging.

## Changelog

:cl:
Readds bailers and seeds to the BMD
Makes them orderable too
Sorts the Dealer Cargo Terminal
Makes it so you get more magazines in bulk cost
:cl:


cheapcents (approved):
Good work

PR #642, opened 2025-01-29
Prevents fire from damaging worn clothes

boy2mantwicethefam:
People can sometimes survive the flames but they often end up almost entirely naked, which is not that fun, as many items can be burnt to ashes.

This is a sort of experimental change purely aimed for QoL rather than consistency because it is still possible to burn items if they are not worn.

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -1990,54 +1990,7 @@ GLOBAL_LIST_EMPTY(selectable_races)
 	if(!CanIgniteMob(H))
 		return TRUE
 	if(H.on_fire)
-		//the fire tries to damage the exposed clothes and items
-		var/list/burning_items = list()
-		var/obscured = H.check_obscured_slots(TRUE)
-		//HEAD//
-
-		if(H.glasses && !(obscured & ITEM_SLOT_EYES))
-			burning_items += H.glasses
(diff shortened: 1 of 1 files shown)


SoreYew:
I feel it is improper to refer to this as QOL due to the fact most instances of fire are combat related.

This balance PR aims to make clothes while worn fireproof, which has significance in combat play patterns, and how players behave when on fire. 

For one example: If a kindred no longer has to care about losing their belt slot, pocket slots, and whatever may be in them, this is a huge mechanical advantage. They are capable of storing their blade to preempt a disarm, having hands free for utilities such as flashlights, etc.

Miraviel:
> People can sometimes survive the flames but they often end up almost entirely naked, which is not that fun, as many items can be burnt to ashes.
> 
> This is a sort of experimental change purely aimed for QoL rather than consistency because it is still possible to burn items if they are not worn.

Please follow the PR template when opening PRs.

Metekillot:
Don't get set on fire if you don't want your clothes to be damaged by fire, or wear fireproof equipment.

Metekillot:
> > People can sometimes survive the flames but they often end up almost entirely naked, which is not that fun, as many items can be burnt to ashes.
> > This is a sort of experimental change purely aimed for QoL rather than consistency because it is still possible to burn items if they are not worn.
> 
> Please follow the PR template when opening PRs.

PR #643, opened 2025-01-29
Recycler no longer deletes itself

boy2mantwicethefam:
## About The Pull Request
Makes the recycler no longer delete itself. This is most notable with the recycler in the restaurant's basement.

## Why It's Good For The Game
It makes the recycler reusable rather than cause it to magically disappear.

## Changelog
:cl:
fix: The recycler in the restaurant's basement will no longer delete itself upon shoving someone in it.
/:cl:

### DIFF
--- a/code/game/machinery/recycler.dm
+++ b/code/game/machinery/recycler.dm
@@ -201,7 +201,6 @@
 	// Instantly lie down, also go unconscious from the pain, before you die.
 	L.Unconscious(100)
 	L.adjustBruteLoss(crush_damage)
-	qdel(src)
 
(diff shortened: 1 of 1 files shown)


Miraviel (changes requested):
It is meant to delete the mob, that is why it made no sense, there was no gibbing back then.

Please follow the PR template, including adding a proper changelog we can paste to the updates channel when we merge PRs.

boy2mantwicethefam:
It is not deleting the mob, it is deleting itself. The mob takes a ton of damage instead. This PR removes that.

PR #659, opened 2025-02-01
Baali can now infiltrate Camarilla/Anarchs/Chantry/Supply

Nokia2Student:
## About The Pull Request

Baali can now infiltrate Camarilla/Anarchs/Chantry/Supply

## Why It's Good For The Game

That lets Baali actually be able to infiltrate different organizations like they could in the lore. Supply is independent/their own faction so Baali should be able to work here as well. Even Caitiff can become a dealer, so why not Baali? 

## Changelog

:cl:
add: Baali can join the round as Scourge, Bruiser, Chantry Archivist, Dealer, Supply Technician
/:cl:

### DIFF
--- a/code/modules/vtmb/jobs/anarchs/bouncer.dm
+++ b/code/modules/vtmb/jobs/anarchs/bouncer.dm
@@ -20,7 +20,7 @@
 
 	display_order = JOB_DISPLAY_ORDER_BRUISER
 	known_contacts = list("Baron")
-	allowed_bloodlines = list("Daughters of Cacophony", "True Brujah", "Brujah", "Nosferatu", "Gangrel", "Toreador", "Tremere", "Malkavian", "Banu Haqim", "Tzimisce", "Caitiff", "Ventrue", "Lasombra", "Gargoyle", "Kiasyd", "Cappadocian", "Ministry")
+	allowed_bloodlines = list("Daughters of Cacophony", "True Brujah", "Brujah", "Nosferatu", "Gangrel", "Toreador", "Tremere", "Malkavian", "Banu Haqim", "Tzimisce", "Caitiff", "Ventrue", "Lasombra", "Gargoyle", "Kiasyd", "Cappadocian", "Ministry", "Baali")
 
 	v_duty = "You are the enforcer of the Anarchs. The baron is always in need of muscle power. Enforce the Traditions - in the anarch way."
 	minimal_masquerade = 2
--- a/code/modules/vtmb/jobs/archivist.dm
+++ b/code/modules/vtmb/jobs/archivist.dm
@@ -21,7 +21,7 @@
 	v_duty = "Keep a census of events and provide information to neonates. Listen to the Regent Carefully. Study blood magic and protect the chantry."
 	minimal_masquerade = 3
 	allowed_species = list("Vampire")
-	allowed_bloodlines = list("Tremere", "Gargoyle")
+	allowed_bloodlines = list("Tremere", "Gargoyle", "Baali")
 	known_contacts = list("Tremere Regent")
 	experience_addition = 15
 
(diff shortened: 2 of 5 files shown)


XanderDox:
I don’t think this will work out. Once they are uncovered as Baali, they will likely still try to join as these jobs and then be killed by everyone that knows they are Baali.

There’s no mechanic to exclude known Baali from taking these roles, I can see a lot of bullshittery and abuse happening as a result. Such as a notorious Baali joining as Scourge and then immediately causing a shit storm for the Tower.

Also, the Scourge are hand picked by the Prince and I find it highly unlikely they wouldn’t screen the clans of those people in some way that was semi reliable.

Nokia2Student:
> I don’t think this will work out. Once they are uncovered as Baali, they will likely still try to join as these jobs and then be killed by everyone that knows they are Baali.
> 
> There’s no mechanic to exclude known Baali from taking these roles, I can see a lot of bullshittery and abuse happening as a result. Such as a notorious Baali joining as Scourge and then immediately causing a shit storm for the Tower.
> 
> Also, the Scourge are hand picked by the Prince and I find it highly unlikely they wouldn’t screen the clans of those people in some way that was semi reliable.

You shouldn't kill Baali because it's Baali. Nothing is stopping you from killing "known?" Baali that are playing as different roles. I did consider about giving them Scourges, but in the end they can always be killed. I'm not sure where did you get hand-picked Scourges bit in lore. Baali made an entire order in Chantry but yet they can't infiltrate Camarilla? I doubt it.

XanderDox:
> > I don’t think this will work out. Once they are uncovered as Baali, they will likely still try to join as these jobs and then be killed by everyone that knows they are Baali.
> 
> > 
> 
> > There’s no mechanic to exclude known Baali from taking these roles, I can see a lot of bullshittery and abuse happening as a result. Such as a notorious Baali joining as Scourge and then immediately causing a shit storm for the Tower.
> 
> > 
> 
> > Also, the Scourge are hand picked by the Prince and I find it highly unlikely they wouldn’t screen the clans of those people in some way that was semi reliable.
> 
> 
> 
> You shouldn't kill Baali because it's Baali. Nothing is stopping you from killing "known?" Baali that are playing as different roles. I did consider about giving them Scourges, but in the end they can always be killed. I'm not sure where did you get hand-picked Scourges bit in lore. Baali made an entire order in Chantry but yet they can't infiltrate Camarilla? I doubt it.

Baali are inherently evil and their existence is illegal, they should be killed when they are identified.

Scourge *are* hand picked by the Prince in WoD lore, they answer directly to him only and carry out his death sentence (and have the authority to unilaterally ash need be).

But yes, in the lore Baali DO infiltrate, but we are not the TTRPG. We're an SS13 server.

The main issue is that our server is HRP and maintains some round to round canonicity. This means if a Baali is uncovered in a role, everyone that finds out about them will ALWAYS know they're Baali, and it'll turn into having to blood hunt them immediately every round. It's not a well thought out or good gameplay loop for _anyone_ involved.

AllCorruptingWyrm:
Of all the things Baali can do, this should not be the step we take as a codebase. The main issue is simply the fact that this doesn't work on a game-ified version of Vampire the Masquerade. It works on pnp because its pure roleplay, but it wont work here. Another thing too is that people aren't going to police their own factions to weed out infernalists unless your Camarilla, but Cmarilla is genuinely filled with drooling idiots who dont know what "balefire" is within the lore. Also, I'm going to actively call into question this PR because I have seen how you act in Round as Gibson, where you went around being an open Baali just mass killing everyone after encouraging everyone to do so. Not a good look.

SpaceSlug2:
> The main issue is that our server is HRP and maintains some round to round canonicity. This means if a Baali is uncovered in a role, everyone that finds out about them will ALWAYS know they're Baali, and it'll turn into having to blood hunt them immediately every round. It's not a well thought out or good gameplay loop for _anyone_ involved.

This server is HRP in name only, nice cope though.

PR #660, opened 2025-02-01
Gargoyle suits

Dusk-a:
## About The Pull Request

Allows gargoyles to wear things in the suit slot.

## Why It's Good For The Game

Their natural protection is outclassed by a $35 trenchcoat and really they should be able to wear most suits without too much trouble.

## Changelog

Gargoyles can wear suit slot items now. Certain suits such as robes will hide their wings in exchange for disallowing their flight.

### DIFF
--- a/code/modules/vtmb/vampire_clane/gargoyle.dm
+++ b/code/modules/vtmb/vampire_clane/gargoyle.dm
@@ -33,7 +33,6 @@
 
 /datum/vampireclane/gargoyle/on_gain(mob/living/carbon/human/H)
 	..()
-	H.dna.species.no_equip = list(ITEM_SLOT_OCLOTHING, ITEM_SLOT_SUITSTORE)
 	H.dna.species.wings_icon = "Gargoyle"
 	H.physiology.armor.melee += 20
 	H.physiology.armor.bullet += 20


Plaguewalker:
I had gargoyles not able to wear suits due to the natural armor potentially stacking, along with them getting fortitude to begin with the original natural armor would've been 30
%. But, realistically you can't properly take away their flight given it is apart of them even if they have a coat, in lore the wings are and would still be visible

Dusk-a:
> I had gargoyles not able to wear suits due to the natural armor potentially stacking with any of the armor available, along with them getting fortitude to begin with. Originally, their natural armor would've been 30%. But, realistically you can't properly take away their flight given it is apart of them even if they have a coat, in lore the wings are and would still be visible even when wearing outerwear

Oh hey, neat to see the original coder here. There was a small little discussion about it in the discord and I don't think armor stacking with the natural protection alongside fortitude is too big of a deal since there's plenty of other people out there running around with Fortitude 5 who _can_ wear proper armor, not to mention some of them having other protection to stack on top like Blood Shield.

But that wasn't the main topic of discussion but rather just the ability to wear more clothes in the first place. Almost every suit available doesn't hide their wings with only the ones that have the "HIDEJUMPSUIT" tag doing so. The only available clothes that have that are robes, the EOD suit which is rarely seen in the first place, and the Bogatyr armor which is essentially exclusive to the Tzimisce from what I've seen. And when their wings are hidden, they can't be used for flight so it's not like you'd suddenly have people hovering around everywhere. If someone would like to change those things so the three people who can even play gargoyle can't hide their wings, feel free to but for now I just want more clothing options available since it's pretty limited and a lot of the modern clothes just don't feel right.

Miraviel (approved):
Code-wise, this works, coats show up and do not hide the wings, robes do hide the wings and disallow flight. The character remains masq-breaching in both cases.

image

Miraviel (changes requested):
Had some talks about this internally with a few people.

This should be good to go if we reworked clothing protection a little. A system where if the armour coming from clothing is higher than the species' innate armour, then use that - instead of the two stacking. (Fortitude would still need to be put on top of this.)

So, ideally, the armor calculation should be: max(species armor, clothing) + Fortitude.

This would ensure we don't accidentally run into an armour stacking problem, now or down the line. If you think you could do this, feel free to add this to the PR. If not, I suggest revisiting this idea in a month or two, when some of our current systems (disciplines and species code) are reworked.

Dusk-a:
I'm not entirely sure that it would be possible to code something like that without some kind of glaring flaw but I've asked some other people who know a bit more about it than I do. Just to discuss possibilities though, what would you think about removing the natural armor for now just so we can get a resolution without waiting a few months and potentially re-add it later down the line with the discipline rework? Because technically in the tabletop Gargoyles don't have any innate protections despite their rock skin _but_ they do get a passive damage soak with a high level of Visceratika so maybe it could tie into the discipline itself when all that stuff does happens.

Dusk-a:
Okay, I think I have a solution. I changed the armor bonus gargoyles get to a brute_mod so armor can't be stacked to total resistance. Feel free to adjust the given number but as of now after all damage calculations, they take 20% less damage.

Miraviel (approved):

PR #663, opened 2025-02-01
Pentex Warehouse major improvements

darkincarnate:
## About The Pull Request

Places a bunch of Pentex keys at the BSD totem and in the mini office. Also does all of the below.

1 - Put a lockable door as a mini-airlock for the warehouse door.
2 - Put a button for the warehouse door on the outside of the warehouse in a mini-security checkpoint, also behind a locked door.
3 - Allow the cars outside Pentex to be unlocked/locked with the Pentex key
4 - Make the cells in the Pentex facility also unlocked/locked by a Pentex key
5 - Allow the fence door leading into the car park to be locked and unlocked by the Pentex key.
6 - Add a garage door in the empty space which is also tied to the security checkpoint
7 - Add extra lattice tile looring just above the all the exposed hell floor tiles so people can enter the facility without going downstairs and not instantly see the flesh of the dark gods covering the floor 
8 - FIX ROOF TILE!
9 - MAKE CARS UNLOCKABLE IN PENTEX!

(Made some changes to admin.txt temporarily on this branch while I was testing, had to set it back to how it was so that's why that file was changed)

## Why It's Good For The Game

Makes the Pentex Warehouse more secure instead of just allowing ANYONE to walk in, alongside adding keys so that more RP can be done around the warehouse, E.G. taking prisoners now that you can unlock and lock the cells.

## Changelog

add: Put a lockable door as a mini-airlock for the warehouse door.
add: Put a button for the warehouse door on the outside of the warehouse in a mini-security checkpoint, also behind a locked door.
add: Allow the cars outside Pentex to be unlocked/locked with the Pentex key
add: Make the cells in the Pentex facility also unlocked/locked by a Pentex key
add: Allow the fence door leading into the car park to be locked and unlocked by the Pentex key.
add: Add a garage door in the empty space which is also tied to the security checkpoint
add: Add extra lattice tile looring just above the all the exposed hell floor tiles so people can enter the facility without going downstairs and not instantly see the flesh of the dark gods covering the floor 
fix: fixes the open roof tiles above the cell stairs

:cl:
add: Added new mechanics or gameplay changes
add: Added more things
del: Removed old things
qol: made something easier to use
balance: rebalanced something
fix: fixed a few things
sound: added/modified/removed audio or sound effects
image: added/modified/removed some icons or images
map: added/modified/removed map content
spellcheck: fixed a few typos
code: changed some code
refactor: refactored some code
config: changed some config setting
admin: messed with admin stuff
/:cl:

image
image
image
image


AllCorruptingWyrm:
I think one of the big issues here is the severe amount of deadspace. Its just, a location, it doesnt feel like a realistic location you can stare at and go, "Yeah people work here" cause remember, the subsidaries of Pentex are very much real companies.

What I'll suggest is an area portal system for the totem. Look at here:
image

AllCorruptingWyrm (commented):
Gave feedback on Discord. Get the mapper role please and @ me there ( @tzula )

PR #665, opened 2025-02-02
Partially(?) Refractors/Moves around some KJ abilities

AllCorruptingWyrm:
## About The Pull Request

I'm probably using the term incorrectly, BUT here we go. KJs are clunky and generally a bit unfun to work with as each ability does separate things, so I took a crack at the code to try and streamline Ghost Flame & Jade Shintai's so there is less "fumbling" around with ability dots. For example, Jade 3 inherits the pass-through tables & dexterity of Jade 2, with the added ability to wall travel as well. Jade 4 inherits both traits and lessens the movement speed, while also making it so you have added flight on top of it, as Jade is at its core a movement-centric Shintai, fumbling around with several dots back and forth doesn't feel good.

I don't know how to code, but I can work on improving things so please, just help me here.

## Why It's Good For The Game

Mentioned above: these two disciplines were extremely clunky and situational; the constant expenditure of chi doesn't help either.

## Changelog

:cl:
tweaks: Drastically improves Jade Shintai's performance as a whole.
Add: Makes it so you can summon the fire katana on ghostflame 5
/:cl:

### DIFF
--- a/code/modules/vtmb/chi_disciplines.dm
+++ b/code/modules/vtmb/chi_disciplines.dm
@@ -204,7 +204,10 @@
 /datum/movespeed_modifier/necroing
 	multiplicative_slowdown = 2
 
-/datum/movespeed_modifier/wall_passing
+/datum/movespeed_modifier/greater_wall_passing
+	multiplicative_slowdown = 2
+
+/datum/movespeed_modifier/lesser_wall_passing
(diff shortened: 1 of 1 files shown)


AllCorruptingWyrm:
this is probably a bad idea.

PR #672, opened 2025-02-04
Debit not Credit

theselfish:
## About The Pull Request

image

## Why It's Good For The Game

See above. No credit is involved.

## Changelog

:cl:
spellcheck: It's a debit card, not a credit card.
/:cl:

### DIFF
--- a/code/modules/vtmb/electronics/atms/atm.dm
+++ b/code/modules/vtmb/electronics/atms/atm.dm
@@ -48,7 +48,7 @@ var/mob/living/carbon/human/H
 		GLOB.bank_account_list += src
 
 /obj/item/vamp/creditcard
-	name = "\improper credit card"
+	name = "\improper debit card"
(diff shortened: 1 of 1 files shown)


code/modules/vtmb/electronics/atms/atm.dm
```diff
@@ -48,7 +48,7 @@ var/mob/living/carbon/human/H
 		GLOB.bank_account_list += src
 
 /obj/item/vamp/creditcard
-	name = "\improper credit card"
+	name = "\improper debit card"
```
Miraviel, line 51:
```suggestion
	name = "debit card"
```

`\improper` is only necessary with objects that have capitalised names, and you need to imply the `\the` text macro on them (`"CPU"` would render as "You throw CPU", `"\improper CPU"` would render as "You throw the CPU")
you can safely remove it here

Miraviel (approved):

PR #676, opened 2025-02-04
Removes True Faith & Numina from Priests

AllCorruptingWyrm:
## About The Pull Request

The following PR removes several tools Priests used to TDM exclusively, with RP being secondary to them. There have been several rounds where players have played a Priest screaming about "Heretics" and meta-targeting every "vampire" hideout because they are a Priest and must destroy "the unholy".

Before anyone says anything. Yes they are getting ahelped and no, there aren't enough staff to police it so a code implementation is being undertaken because everyone abused it. Yes, I've ahelped, its getting cracked down on.

## Why It's Good For The Game

Abilities should not be something you get for existing at round start; you should work for them (some, more than others i.e xp/money/etc), with that said, Priest has became reviled due to its /tg/ players treating the Priest role like they are Boris from terry.

Also another thing, true faith is supposed to be rare. Not everyone gets it, and neither should a role that anyone can be get it as people fundamentally don't understand the significance it has.

## Changelog
Removes numina/true faith from the priest.
Tweaks priest objectives.

### DIFF
--- a/code/modules/vtmb/jobs/priest.dm
+++ b/code/modules/vtmb/jobs/priest.dm
@@ -20,8 +20,7 @@
 	allowed_species = list("Human")
 	minimal_generation = 13
 
-	duty = "Be the shepherd of the flock in San Francisco, lead them to salvation, piety and righteousness."
-	v_duty = "The mortals believe you to be their savior. The kindred look at you with suspicion. Yours is the charge of this church."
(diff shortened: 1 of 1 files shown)


Metekillot:
Making an antagonistic role miserable to play against ludicrously overturned vampires isn't a good solution. If you want to hack off late party TDM bullshit, just disable late party. We're getting rid of midround style auto spawning antags.

PR #683, opened 2025-02-05
Orderable vitae packs in cargo

K4rlox:
## About The Pull Request

image
vitae packs can now be ordered in cargo, same way and price as blood packs

## Why It's Good For The Game

previously, the only way to get vitae packs was to get it thru several places they are MAPPED IN
and when they were out, you couldnt get more

now you can buy them, similar to blood packs

## Changelog

:cl:
add: added elder blood pack subtype because people were mapping them in previously
add: elder vitae pack in cargo
add: vitae pack in cargo
/:cl:

### DIFF
--- a/code/modules/wod13/cargo.dm
+++ b/code/modules/wod13/cargo.dm
@@ -38,6 +38,20 @@
 	contains = list(/obj/item/drinkable_bloodpack/elite, /obj/item/drinkable_bloodpack/elite, /obj/item/drinkable_bloodpack/elite, /obj/item/drinkable_bloodpack/elite, /obj/item/drinkable_bloodpack/elite)
 	crate_name = "blood crate"
 
+/datum/supply_pack/vampire/vitaepack
+	name = "Vitae Pack"
+	desc = "Contains 5 vitae packs."
+	cost = 100
+	contains = list(/obj/item/drinkable_bloodpack/vitae, /obj/item/drinkable_bloodpack/vitae, /obj/item/drinkable_bloodpack/vitae, /obj/item/drinkable_bloodpack/vitae, /obj/item/drinkable_bloodpack/vitae)
+	crate_name = "vitae crate"
+
+/datum/supply_pack/vampire/vitaepack_elder
+	name = "Elder Vitae Pack"
+	desc = "Contains 5 elder vitae packs."
+	cost = 300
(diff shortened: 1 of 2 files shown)


AllCorruptingWyrm:
no.

theXDkidoflol:
i forget but can't you get these by using the blood harvesting things in the clinic on a vampire

K4rlox:
> Should be incredibly expensive or not a thing at all. BP is already an excessively common resource and entire gameplay systems are largely circumvented by vampires just sucking down an effectively infinite supply of blood packs every round. Make people actually have to put fangs in necks.

heres a easy response: when you code it

this is a pr to add vitae packs and not removing blood packs, please refer this idea to suggestions channel in the discord

so that your suggestion doesnt hit a wall where nobody will take it

XanderDox:
This would lead to horrific balancing issues that already exist with the rich kindred buying 20 orders of elite blood.

No matter how expensive this is, people will make enough to mass purchase it and that will ruin the bloodpoint balance across the board.

As well, it implies there is a Vitae factory out there harvesting Kindred or Revenants en masse for Vitae, which is not blood but the literal soul stuff of vampires. Vitae is also meant to expire shortly after leaving the body, I don't think it would reasonable survive shipping.

Miraviel (changes requested):
Personally, I'd say no to this PR, I think we need way less source of bagged blood, both mapped in and orderable - vampires should be more responsible with their blood and not be infinite murder machines. The current ones priced at $300 are already an issue. This contributes to it.

But I'll leave it open for other maintainers to look at it too

hex37:
Vitae packs in general are not super lore friendly and I think they should be removed. Independent ghouls shouldn't get a free pass to avoid having to negotiate for a sippy. Figuring that shit out is the whole point of being an indpendent ghoul. I think it shouldn't be a thing.

wgtjunior743:
i would say that vitae is not that bad but Elder Vitae would create a very big issue, very low price for a very op blood bag that always get meta rushed 5 mins into the game for the lucky person to get, it will drag out any fight and give an unfair advantage against werewolves or kueijin for such a op blood bag also it makes no sense that ELDER blood out of all things would be sold in BULK like paint and with such a low price.

K4rlox:
> Personally, I'd say no to this PR, I think we need way fewer sources of bagged blood, both mapped in and orderable - vampires should be more responsible with their blood and not be infinite murder machines. The current ones priced at $300 are already an issue. This contributes to it.
> 
> But I'll leave it open for other maintainers to look at it too

i think you pressed the wrong button?
you requested changes...?

Metekillot:
Too gamey, too against the theme, creeps into the territory of blood alchemy that should be a way for independent ghouls or cannibalistic vampires to get vitae packs. However, if you want to make things a little less arcade-y, make the blood pack factories be unusable on NPCs.

PR #685, opened 2025-02-05
Adds Fire Suppression to Church

XanderDox:
## About The Pull Request

Adds Fire Suppression to the Church.
image

## Why It's Good For The Game

The Church is regularly set on fire, causing the priest to lose their entire building and base almost every other round. This won't stop the building from burning entirely, and committed people will still be able to wreck it if they want to, but this should help priests and their church last a little longer.

## Changelog

:cl:
add: Added Fire Suppression to Church
:cl:

### DIFF
--- a/code/modules/wod13/ambient.dm
+++ b/code/modules/wod13/ambient.dm
@@ -272,6 +272,7 @@
 	music = /datum/vampiremusic/church
 	ambience_index = AMBIENCE_INTERIOR
 	upper = FALSE
+	fire_controled = TRUE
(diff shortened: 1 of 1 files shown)


XanderDox:
Requested in this Suggestion: https://discord.com/channels/1293720761739247769/1336607867544797215

Metekillot:
We need to let people sabotage fire suppression, in the future. As it is now, fire being utterly unstoppable without it will need mitigation. Approved

Metekillot (approved):

PR #687, opened 2025-02-06
QOL, Card is usuable direclty in most shops.

wgtjunior743:
## About The Pull Request

I had this idea for a while just took a while to start, basically you can insert your card into any shop besides the illegal products, asking for your pin and then it will have your bank money , that you can use to buy the items in the shop, the shop will store the card while you use it, you can alt click to withdraw it.

I am not sure on the idea that you can't use cards to buy illegal stuff in the sense of how easy is to a illegal business to fake what they are selling in the USA and ppl being able to do that with no issues, i know stuff about the silk road but this is much earlier.

Buying stuff:

buying_stuff_1

buying_removing

Black market shops:
black_market

## Why It's Good For The Game

I always felt like we barely used the card in the game, atleast for me and most of the people is to use the card to withdraw all our money and then use to do whatever, now atleast you got a reason to keep some money in your card so you can easily buy anything non illegal.

## Changelog

:cl:
fix: Minor message that used points instead of money.
qol: You can know use your card in all the shops besides the illegal ones, BM and laundrymat.
/:cl:

### DIFF
--- a/code/modules/mining/machine_vending.dm
+++ b/code/modules/mining/machine_vending.dm
@@ -11,7 +11,9 @@
 	var/mob/living/carbon/human/npc/my_owner
 	var/owner_needed = TRUE
 	var/obj/item/card/id/inserted_id
+	var/obj/item/vamp/creditcard/creditcard_instance = new /obj/item/vamp/creditcard()
 	var/points = 0
+
 	var/list/prize_list = list( //if you add something to this, please, for the love of god, sort it by price/type. use tabs and not spaces.
 		new /datum/data/mining_equipment("1 Marker Beacon",				/obj/item/stack/marker_beacon,										10),
 		new /datum/data/mining_equipment("10 Marker Beacons",			/obj/item/stack/marker_beacon/ten,									100),
@@ -132,7 +134,8 @@
 	.["user"]["points"] = points
 	.["user"]["name"] = "[user.name]"
 	.["user"]["job"] = "[user.mind.assigned_role]"
-
+	if(creditcard_instance.inside_shop)
+		.["user"]["job"] = "(Your card is activated! Use Alt+Click to remove the card.)"
 
 /obj/machinery/mineral/equipment_vendor/ui_act(action, params)
 	. = ..()
@@ -163,10 +166,14 @@
 				flick(icon_deny, src)
 				return
 			if(prize.cost > points)
-				to_chat(usr, "<span class='alert'>Error: Insufficient points for [prize.equipment_name]!</span>")
+				to_chat(usr, "<span class='alert'>Error: Insufficient money for [prize.equipment_name]!</span>")
 				flick(icon_deny, src)
 				return
-			points -= prize.cost
+			if(creditcard_instance.inside_shop)
+				points -= prize.cost
+				creditcard_instance.account.balance -= prize.cost
+			else
+				points -= prize.cost
 			to_chat(usr, "<span class='notice'>[src] clanks to life briefly before vending [prize.equipment_name]!</span>")
 			new prize.equipment_path(loc)
 			SSblackbox.record_feedback("nested tally", "mining_equipment_bought", 1, list("[type]", "[prize.equipment_path]"))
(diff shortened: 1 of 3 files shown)


Miraviel (changes requested):
Debit cards are underutilised, that is indeed a problem. However, I would not add paying with cards in shops (and later, the warehouse, if we wanted to be consistent), simply because the game is set in 2005. Paying with card everywhere was barely a dream back then.

I know sometimes mechanics are more important than the lore/atmosphere, but I think this is one of those things that would take away from it. If cards were used like this, we could just go back to ID cards and slowly start reverting to default ss13 features in the name of comfort.

wgtjunior743:
> Debit cards are underutilised, that is indeed a problem. However, I would not add paying with cards in shops (and later, the warehouse, if we wanted to be consistent), simply because the game is set in 2005. Paying with card everywhere was barely a dream back then.
> 
> I know sometimes mechanics are more important than the lore/atmosphere, but I think this is one of those things that would take away from it. If cards were used like this, we could just go back to ID cards and slowly start reverting to default ss13 features in the name of comfort.

i would have to disagree with that, after a couple of searches in the internet about specially debit cards usage in 2000-2006 and then digging up further upon getting some articles from chat gpt such as :  https://www.federalreserve.gov/pubs/feds/2006/200616/200616pap.pdf?utm_source=chatgpt.com and https://www.kansascityfed.org/Research/documents/6675/GuideATM_DebitCardIndustry_2016.pdf?utm_source=chatgpt.com , people were shifting from using ATM with their cards to the usage of debit cards directly and in a few years it would what people would always do,

At the moment the debit card is by far not used and could even be called useless, if interacting directly with an atm to withdraw our money and make transactions without a card, people would not even care, the point is that the usage of cards and debit cards were real back then, they started earlier as 1960s in the USA, San Francisco is not a small city and the game is in the USA both a first world country and a place with innovation , if it was a game situated in a third-world country i would agree with you or a place like Japan or Germany besides being first-worlds have a general heavy usage of money rather than cards to use in shops.

wgtjunior743:
> Debit cards are underutilised, that is indeed a problem. However, I would not add paying with cards in shops (and later, the warehouse, if we wanted to be consistent), simply because the game is set in 2005. Paying with card everywhere was barely a dream back then.
> 
> I know sometimes mechanics are more important than the lore/atmosphere, but I think this is one of those things that would take away from it. If cards were used like this, we could just go back to ID cards and slowly start reverting to default ss13 features in the name of comfort.

And please don't request changes on my PRs unless you are improving the code or you are maintainer yourself.
There is the suggest changes option and that is what people do.

Miraviel:
> And please don't request changes on my PRs unless you are improving the code/fixing an issue or you are a maintainer yourself. There is the suggest changes option and that is what people do.

Only maintainers and above can do that. I use request changes as a blocking review, because there is no other option on Github besides Comment (does not block) Approve (approval) and Request Changes.

wgtjunior743:
> > And please don't request changes on my PRs unless you are improving the code/fixing an issue or you are a maintainer yourself. There is the suggest changes option and that is what people do.
> 
> Only maintainers and above can do this, I am a maintainer myself. I use request changes as a blocking review, because there is no other option on Github besides Comment (does not block) Approve (approval) and Request Changes.

an example of a suggested changes : 
image

The github site lacks some features, that type of suggestion needs to be in spaces like vscode

Miraviel:
> an example of a suggested changes : image
> 
> The github site lacks some features, that type of suggestion needs to be in spaces like vscode

I am aware of this - but I have no code suggestion to make in this situation, as I would prefer if cards just did not work in shops, in general, to facilitate the usage of cash more, which is the entire PR. Again, this is my own, individual opinion, so if the other maintainers take other position, I'll swap it to approved and let them merge it.

XeonMations:
I don't really see a reason for this, since in 2005, as @Miraviel said, credit card purchases were scarce. Not to mention the various issues that might come up with not having money on your person, when someone might attempt to rob you.

🔑

PR #715, opened 2025-02-08
removed several odd bloodpacks

xxxnoscopexx:
removed the overpowered bloodpacks because 40 point bloodpacks seem kinda silly

## About The Pull Request
removes several bloodbags that are varedited 

## Why It's Good For The Game
gets rid of overpowered hidden bloodbags

## Changelog
removes several bloodbags that are varedited and sneakly hidden around the map

:cl:
removes several bloodbags that are varedited and sneakly hidden around the map
/:cl:


Miraviel (changes requested):
lol, lmao even, nice catch

you need to rebase the map, your map doesn't seem to have the changes of #710

PR #717, opened 2025-02-08
Fixes Accidental Deletion of Old Clan Stairs/TransferPoint from Pentex Remap

XanderDox:
## About The Pull Request

During the Pentex remap, the Old Clan's stairs to their basement and corresponding transfer point were both deleted accidentally. This restores them.

## Changelog

:cl:
map: Re-adds stairs and a transfer point to the Old Clan's manor to access their basement.
:cl:


XanderDox:
New stairs.

https://github.com/user-attachments/assets/b73ba6b9-e2c0-4116-bb2b-2b49286a8751

Miraviel (changes requested):
Need a rebase because of #721 , but you can just do that manually (rebase the code so you get the new police chief office door, then replace the broken one with it)

Also make sure the stairs are usable across the map, you changed every single stairs according to the mapdiffbot

XanderDox:
I simply don't have the energy to go back through this after having to spend over an hour on it last night alone because of how broken our stairs are in the first place. Someone else will have to fix this.

XanderDox:
I changed my mind >.>

Here is the updated fix. I checked multiple staircases and found no issues, I did not actually use stairs, I reskinned a single carpet to look like a staircase because our stair object is broken as hell and nobody has fixed it.

I went through like 100 of the carpet instances to make sure no others looked like a staircase. We should be good.

Miraviel (approved):

PR #504, opened 2025-01-09
Malkavian Madness Network changes and updates to it.

wgtjunior743:
## About The Pull Request

- FIrst of all some changes for Malkavian directly : Talking into the hivemind requires dementation 3 or more to simulate the difficulty of being able to talk there.

- Cooldown to talk increased but it scales off your dementation level lowering it, if your Dementation is lvl 1 it will be 35 seconds,  if it is level 5 the CD will be now : 15 seconds.

- Dementation levels also now decrease the chance of leaking your conversation into the madness network on the offset of increasing the chance of leaking your conversation to 35% on level 1 of dementation and 15% on level 5 of dementation.

- Lower gen slightly increase your font size on the madness network to simulate olders vampires influence in there.

- Radios Hallucinations are now stylized to be like it is inside the madness network for the user.

- Encrypt System added to the Madness network scaling of your auspex, with one level of auspex there will be a 50% chance that anything told inside the madness network will be modified to a random word, messing up what has been told, it tries to kinda keep the message not way to mess up with the different templates of the messages.

- Level 5 Auspex will have a 10% chance of scrambling information on the network so keep yourselves on your toes!

Examples:

(it won't show the auspex lvl on the chat ingame)

Auspex 1 : 
auspex_1_dnv

Auspex 5:
auspex_5

The font size is random but just abit to simulate if someone was talking on it.
Radio Hallucination:

radio_hallucination

## Why It's Good For The Game

Helps with the issue of the madness networking becoming a chat and not actually a madness network, it will now confuse more people with the previous radio chats being inside of it and now stuff that is leaked will be changed up as well.

## Changelog
:cl:
add: New stuff for the madness network!
/:cl:

PR #722, opened 2025-02-09
update hallucinations

buffyuwu:
## About The Pull Request
adds hallucination texts loosely gathered from the game via [the gamespot forums](https://gamefaqs.gamespot.com/pc/914819-vampire-the-masquerade-bloodlines/answers/184545-malkavian-whispers-list)

Also, disables radio hallucinations from happening by setting the is_radio var to false. I didn't remove the rest of the radio hallucination code because theres a pending pr that refactors it. this just disables them for now

### Some new hallucinations (on live currently these three messages would either be HELP! or The prince is a traitor!)
image

## Why It's Good For The Game

more hallucinations good.

the other part of this PR that was removed will be converted into a code bounty

## Changelog

:cl:
add: Added many more hallucination messages
/:cl:

### DIFF
--- a/code/modules/surgery/organs/tongue.dm
+++ b/code/modules/surgery/organs/tongue.dm
@@ -10,8 +10,10 @@
 	attack_verb_simple = list("lick", "slobber", "slap", "french", "tongue")
 	var/list/languages_possible
 	var/say_mod = null
+
 	/// Whether the owner of this tongue can taste anything. Being set to FALSE will mean no taste feedback will be provided.
 	var/sense_of_taste = TRUE
+
 	var/taste_sensitivity = 15 // lower is more sensitive.
 	var/modifies_speech = FALSE
 	var/static/list/languages_possible_base = typecacheof(list(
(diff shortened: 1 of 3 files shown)


theselfish:
I like it, but not all Malkavians are in the schizo-affective vein. This would also ruin any attempts at a Malkavian who isn't obvious.

boy2mantwicethefam:
It's amazing though I do have some suggestions. I do recommend making it a toggleable ability, perhaps an action granted by the tongue that allows toggling a variable on the tongue that will check whether it will do regular speech or malkavian speech?

code/modules/surgery/organs/tongue.dm
```diff
@@ -281,6 +281,58 @@
 	icon_state = "tongueplasma"
 	modifies_speech = FALSE
 
+/obj/item/organ/tongue/malkavian
+	name = "tongue"
+	desc = "A fleshy muscle mostly used for telling the truth."
+	icon_state = "tonguenormal"
+	modifies_speech = TRUE
+	var/chattering = FALSE
```
boy2mantwicethefam, line 289: Is making chattering bone noises when talking intentional? If not you can remove var/chattering, var/phomeme_type, var/list/phomeme_types and var/static/list/languages_possible_skeleton

code/modules/surgery/organs/tongue.dm
```diff
@@ -281,6 +281,58 @@
...
+		/datum/language/arabic,
+		/datum/language/german,
+		/datum/language/hebrew,
+		/datum/language/japanese,
+		/datum/language/cantonese,
+		/datum/language/greek
+	))
+
+/obj/item/organ/tongue/malkavian/Initialize()
+	. = ..()
```
boy2mantwicethefam, line 311: You can delete this Initialize()

code/modules/surgery/organs/tongue.dm
boy2mantwicethefam: And also delete the chattering stuff in handle_speech

buffyuwu:
processing changes based on feedback

buffyuwu:
> It does look really neat but I think overall I'd prefer if it was something that randomly happened (maybe triggers on messages sent through the madness network) rather than a 100% all the time thing

making it toggleable, which will roundstart disabled

wgtjunior743:
Very nice, once the discipline freeze is done,the radio will be back with this PR : https://github.com/WorldOfDarknessXIII/World-of-Darkness-13/pull/504, it will introduce the radio chat into the malk hivemind with other change to that as well.

Miraviel (changes requested):
Funny feature, but the code is a mess. You declared a proc, and then used verb syntax for it.

You don't need the item either, people will just find a way to cut it out and replant it into something else. I think you can make it into an action and assign it to malkavians upon spawning in, you should be able to do it when they receive their hivemind (though I'm not sure if they already have a client by then)

wgtjunior743:
> Funny feature, but the code is a mess. You declared a proc, and then used verb syntax for it.
> 
> You don't need the item either, people will just find a way to cut it out and replant it into something else. I think you can make it into an action and assign it to malkavians upon spawning in, you should be able to do it when they receive their hivemind (though I'm not sure if they already have a client by then)

they do got a client, if there is a H(mob/living/carbon/human), there is a client inside of it

XanderDox:
@buffyuwu , will you be closing this PR now that #754 is up?

Miraviel (approved):

PR #725, opened 2025-02-09
Adds a typing indicator to the me command, small say refactor, crinos speech bubble

Miraviel:
## About The Pull Request

Updated our say command a little, here are the changes:
- Added the speech bubble to the me and whisper commands
- Added a line ending (...), (!) or (?) bubble that lingers for 2 seconds after you sent a verbal message, depending on the last character of the message
- Refactored how the typing indicator works
- Added a limitation of 512 characters to the say, me, and whisper commands
- Added a custom bubble for the Crinos form, if the Garou players like this, we can have one for the Glabro and Lupus forms as well (we just need sprites, it's a one-line code change)
- Removed the input_say() verb, it was pointless

NPCs currently do not use this system.

## Why It's Good For The Game

We are supposedly a roleplay server, so small features like this help with immersion. Now you can see if someone is writing a lengthy emote.

This can be later expanded to `atom_say()` that I also intend to port over.

The only concern I have that I might have missed a case where the bubble lingers, but that might be a trial & error thing to find.

## Changelog

- Added the speech bubble to the me and whisper commands.
- Added a line ending (...), (!) or (?) bubble that lingers for 2 seconds after you sent a verbal message, depending on the last character of the message.
- Added a limitation of 512 characters to the say, me, and whisper commands.

### DIFF
--- a/code/modules/mob/mob_say.dm
+++ b/code/modules/mob/mob_say.dm
@@ -55,9 +55,9 @@ GLOBAL_LIST_EMPTY(typing_indicator)
 		to_chat(usr, "<span class='danger'>Speech is currently admin-disabled.</span>")
 		return
 
-	set_typing_indicator(TRUE)
+	set_typing_indicator(TRUE, TRUE)
 	var/message = input("What are you trying to emote? (A maximum of [MAX_BROADCAST_LEN] characters])") as text|null
-	set_typing_indicator(FALSE)
+	set_typing_indicator(FALSE, TRUE)
 
 	if(!message)
 		return
--- a/code/modules/mob/mob_say.dm
+++ b/code/modules/mob/mob_say.dm
@@ -18,7 +18,6 @@ GLOBAL_LIST_EMPTY(typing_indicator)
 
 	if(!message)
 		return
-	message = message_clean(message)
 
 	say(message)
 
@@ -38,7 +37,6 @@ GLOBAL_LIST_EMPTY(typing_indicator)
 
 	if(!message)
 		return
-	message = message_clean(message)
 
 	whisper(message)
 
--- a/code/modules/mob/mob_say.dm
+++ b/code/modules/mob/mob_say.dm
@@ -13,10 +13,10 @@ GLOBAL_LIST_EMPTY(typing_indicator)
 		return
 
 	set_typing_indicator(TRUE)
-	var/message = input("What are you trying to say? (A maximum of [MAX_BROADCAST_LEN] characters]") as text|null
+	var/message = input("What are you trying to say?") as text|null
 	set_typing_indicator(FALSE)
 
-	if(!message)
+	if(!message || message_max_length_check(message))
 		return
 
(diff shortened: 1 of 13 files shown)


code/modules/mob/mob_say.dm
```diff
@@ -1,52 +1,74 @@
...
 	set name = "Whisper"
 	set category = "IC"
+
+	set_typing_indicator(FALSE)
 	if(GLOB.say_disabled)	//This is here to try to identify lag problems
 		to_chat(usr, "<span class='danger'>Speech is currently admin-disabled.</span>")
 		return
+
+	set_typing_indicator(TRUE)
+	var/message = input("What are you trying to whisper? (A maximum of [MAX_BROADCAST_LEN] characters]") as text|null
```
TheCarnalest, line 35: Having the maximum char limit here feels like clutter. Could you just add a check that alerts people if their message has been cut off and gives them the original one back?
```suggestion
	var/message = input("What are you trying to whisper?") as text|null
```

code/modules/mob/mob_say.dm
```diff
@@ -1,52 +1,74 @@
...
 	set category = "IC"
+
+	// This is to avoid multiple instancing
+	set_typing_indicator(FALSE)
 	if(GLOB.say_disabled)	//This is here to try to identify lag problems
 		to_chat(usr, "<span class='danger'>Speech is currently admin-disabled.</span>")
 		return
+
+	set_typing_indicator(TRUE)
+	var/message = input("What are you trying to say? (A maximum of [MAX_BROADCAST_LEN] characters]") as text|null
```
TheCarnalest, line 16: Same as above

code/modules/mob/mob_say.dm
```diff
@@ -1,52 +1,74 @@
...
 	set category = "IC"
 
+	set_typing_indicator(FALSE, TRUE)
 	if(GLOB.say_disabled)	//This is here to try to identify lag problems
 		to_chat(usr, "<span class='danger'>Speech is currently admin-disabled.</span>")
 		return
 
-	message = trim(copytext_char(sanitize(message), 1, MAX_MESSAGE_LEN))
+	set_typing_indicator(TRUE, TRUE)
+	var/message = input("What are you trying to emote? (A maximum of [MAX_BROADCAST_LEN] characters])") as text|null
```
TheCarnalest, line 57: Same as above

TheCarnalest (changes requested):
Just a nitpick and it's good to go.

XeonMations (changes requested):
Make sure to stop malkavian code from picking up the says. (See chameleonproj.dm)

Miraviel:
image

XeonMations (approved):

Agateophobia (approved):
LGTM

Agateophobia (changes requested):
After testing it for the third time, Crinos form doesn't seem to use a different speech bubble.

PR #733, opened 2025-02-10
Gives Everyone 3 Quirk Points

XanderDox:
## About The Pull Request

Gives ~~Garou~~ all players 3 quirk points to work with at character creation/change appearance, the same as humans. ~~Presently, Garou are almost entirely unable to take quirks that are very beneficial for roleplay (dancer/languages) without making their character essentially disabled, as 90% of the negative quirks are vampire restricted.~~

## Why It's Good For The Game

~~Garou~~ Everyone deserves to speak other languages without missing an arm or being blind or walking at snail speed.

## Changelog

:cl:
add: Gives everyone 3 quirk point balance at character creation/change.
:cl:

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -1576,7 +1576,7 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 
 /datum/preferences/proc/GetQuirkBalance()
 	var/bal = 0
-	if(pref_species.name == "Human")
+	if(pref_species.name == "Human" || "Werewolf")
 		bal = 3
 	for(var/V in all_quirks)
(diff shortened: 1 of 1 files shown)


code/modules/client/preferences.dm
```diff
@@ -1576,7 +1576,7 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 
 /datum/preferences/proc/GetQuirkBalance()
 	var/bal = 0
-	if(pref_species.name == "Human")
+	if(pref_species.name == "Human" || "Werewolf")
```
Metekillot, line 1579: Just remove the if and the indent to give everyone 3 quirk points

Metekillot (changes requested):

PR #738, opened 2025-02-11
Silver bullet nerf

DiscordWizard:
## About The Pull Request

This PR tidies silver bullet code and creates a proper status effect to apply slowdown on anything that silver is bane of. I plan to make a proper Bane system in the future but this works for now.

This PR also removes the large amount of free brute damage silver bullets dealt to werewolves while keeping the guaranteed clone damage. Considering werewolves do not have any of the human code that allows more damage resistance, this is to stop werewolves from disappearing the second they are shot with a silver bullet. It also normalises silver bullet damage so some bullets aren't better by default than their normal counterparts.

## Why It's Good For The Game

It makes the code better. It makes damage by silver bullets actually care about armor and resistances.

## Changelog

:cl:
balance: removes extra brute damage of silver against werewolves
fix: code tidying
/:cl:


Metekillot (changes requested):
Please remove the extra brute damage, the impossible-to-heal clone damage is more than enough. Cheers on removing that ridiculous hardstun

Metekillot (changes requested):
Oh, one MORRRREEEEE thing: 

Please see the following discord post about a new policy regarding file organization 

https://discord.com/channels/1293720761739247769/1294402575826812939/1338606607998783612

Miraviel:
this has a dme conflict

@Metekillot are you sure you want all brute damage to be gone? it's a bit weird to be hit by literal bullets and not get a single booboo from them
even just a tiny amount would work so people get properly bloodied (and hopefully bleed in the future when our medical system is less basic)

Nyvrem:
All the silver ammo types are children of the regular ammo types they're based on. Removing the damage from the silver types will mean they instead use the damage value of regular ammo. Which means:

Silver 5.56mm will go from 35 to 45. Also this PR gives it 10 extra AP on top of that.
Silver 9mm will go from 23 to 18.
Silver .45 ACP will go from 25 to 20.
Silver .44 will go from 40 and 35. (The AP doesn't change. IDK why the child assigns it the same AP as the parent but removing that is good.)

They will still be BULLETS, they just won't do extra brute to garou on top of being bullets. Only extra clone damage.

AllCorruptingWyrm:
I wouldnt give silver 5.56 more armor pen

Miraviel (approved):

PR #760, opened 2025-02-15
Adds Keeper/Harpy Badges and Keys to Prince.

XanderDox:
## About The Pull Request

In the same vein as my Primogen Badge PR, this PR adds two new badges but spawns them with the Prince:

- Topaz Keeper of Elysium's Badge
- Amethyst Harpy's Badge

In addition, the PR creates two new keys that also spawn with the Prince:

- Harpy's Keys (Camarilla, MilleniumCommon)
- Keeper's Keys (Camarilla, MilleniumCommon, Archive)

Lastly, the PR fixes the issues with side-profile icons that showed up on shoulders, by stopping them from showing. These badges are very small and should not be visible from a side profile anyways.

ameythyston
badges
topaz on
inventory sprites

## Why It's Good For The Game

Players have been requesting these two roles for quite awhile, but staff perspective has been that they do not need to be full independent roles of their own. To bridge this gap, the Prince can now assign one of his court (one without a job or one already with a job) to be Harpy and Keeper, and carry the privileges of those titles.

## Changelog

:cl:
add: Added two new badges, Harpy Badge and Keeper Badge, for Prince distribution.
add: Added two new keys, Harpy Keys and Keeper Keys, for Prince distribution.
fix: Side profile of all Camarilla Badges showing on shoulders.
sprites: Adds sprites for two new badges.
:cl:


CMDR-Gungnir:
In a similar vein to this, could we have extra P25's spawn in the Prince's office, in case the Tower's bringing people into the team? Or would that be better as a separate PR?

squidqueen:
I still think that Harpies should be jobs, Anarchs already get their version of harpies as jobs, and more social roles in the cam vs fragging is a good thing I think

Miraviel:
> staff perspective has been that they do not need to be full independent roles of their own

can you quote where this decision was made? the harpy role is definitely something that would be useful as a job, having a shiny sticker that is only visible on examination sounds like a waste for such a vital role in the cam

XanderDox:
> > staff perspective has been that they do not need to be full independent roles of their own
> 
> can you quote where this decision was made? the harpy role is definitely something that would be useful as a job, having a shiny sticker that is only visible on examination sounds like a waste for such a vital role in the cam

It was mostly HomuHomu saying he did not see why positions that could be assigned as duties to Hounds or the Seneschal should be independent jobs. Not necessarily a decision, but if I have an inkling that administration is against something I'm not going to waste time PRing it.

>> shiny sticker
rood

squidqueen:
Administration don't control the code direction, right?

silencer-pl:
Harpies are absolutely fully defined roles that deserve their own slot and in the future, their own social space to run (though currently they can latch on to multiple Camarilla-aligned spots already available.

The Camarilla, per their own description, is an organization where words are supposed to carry more worth than physical power and yet there are way more "combat" job slots available. Relegating the Harpy in particular to a pin (which needs a Prince, needs a player to waste time arranging getting the pin or literally making semi-ooc agreements which both seem to go against the spirit of an RP focused server) seems to go against this. Let the Seneshal truly be the Princes second and Harpy be the primary social role. In fact there should be more Camarilla aligned non-Prince roles like this that are focusing on social, not combat interaction.

The only reason I'm not suggesting the same for the Keeper is that right now there isn't really an Elysium to keep.

PR #824, opened 2025-02-19
ancillae can be primogen

hex37:
## About The Pull Request

Makes the youngest generation that can be a primogen 10

## Why It's Good For The Game

Allows more variety in certain job positions. According to SquidQueen (and no other sources because I didn't look), the HIGH REGENT of the Tremere in new york is 10th gen

<details>

it compiled,,,

</details>

## Changelog

:cl:
balance: Makes the youngest generation that can be a primogen 10
/:cl:


code/modules/vtmb/jobs/primogen.dm
```diff
@@ -18,7 +18,7 @@
 
 	allowed_species = list("Vampire")
 	allowed_bloodlines = list("Malkavian")
-	minimal_generation = 7	//Uncomment when players get exp enough
+	minimal_generation = 10	//Uncomment when players get exp enough
```
Miraviel, line 21: remove these comments please
```suggestion
	minimal_generation = 10
```

Miraviel (changes requested):

squidqueen:
https://whitewolf.fandom.com/wiki/Aisling_Sturbridge 10th gen

Miraviel (dismissed):

XanderDox:
May be relevant to mention that Regents are not technically Primogen and have their own separate selection process and expectations.

Because the Pyramid/Tremere are hierarchical, the Regent defacto and dejure represents the clan, thus there being no Tremere Primogen in their place.

silencer-pl:
> given that Primogen are by appointment of the Prince.

The Primogen are not appointed by the Prince. The Prince can, through political meddling and favor currying make the Council his effective puppet, but the Clans each pick their own Primogen via internal processes.

The revised Camarilla corebook says that each clan picks their representive for the council and that the honor is typically bestowed upon the oldest (as in by actual age) Vampire in the city, but is not limited to that and can very much happen because someone says so.

Each clan in fact has their own internal process of choosing their Primogen. Most have some sort of democracy, but the Ventrue for instance can just put someone on.

Either way, Primogen shouldn't really be generation locked as that does not represent the requirements for the post.

XeonMations (approved):

PR #893, opened 2025-02-24
Makes the cost of raising your Dharma level scale with level (also, a non-critical but substantial bug fix)

faaaay:
## About The Pull Request

Previously, Kuei-jin had to spend 20 XP for each Dharma level. This PR changes that cost to 5 XP per current level to a maximum of 20, so the costs are 5, 10, 15, 20, 20.
Also fixes the way Kuei-jin max HP is calculated, because this didn't properly take both physique and dharma into account.

## Why It's Good For The Game

The benefits of Dharma levels 2-4 are limited to +25 HP, and 20XP is just a crazy cost for that at the moment - Physique does that and more! Only 5 and 6 provide any other benefit, those being other ways to draw chi (very useful) and an increase to your maximum resources.

<details>

image
I have 935 XP.
image
I now have 925 XP.
image
The above image was taken with 6 physique and Dharma 3.
image
This one was taken with 5 physique and Dharma 6.
</details>

## Changelog

:cl:
balance: kuei-jin only need to pay 5, 10, and 15 XP for the first three dharma rises
fix: kuei-jin should have the correct HP for their level and stats
/:cl:

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -2292,10 +2292,10 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 					enlightenment = !enlightenment
 
 				if("dharmarise")
-					if ((true_experience < 20) || (dharma_level >= 6) || !(pref_species.id == "kuei-jin"))
+					if ((true_experience < min((dharma_level * 5), 20)) || (dharma_level >= 6) || !(pref_species.id == "kuei-jin"))
 						return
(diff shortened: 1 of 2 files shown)


hugzar:
Peak

hex37:
Might be necessary, considering that xp will be a fixed number in the future.

XeonMations (approved):

PR #942, opened 2025-03-13
More Bloodlines for Baron

gr4ts:
Includes Baali, Caitiff and Banu Haqim as allowed Barons. Caitiff should be allowed to be Barons as it's the most feasible position of 'power' one could obtain, disallowing an actual Prince as I consider the one example of that, the Prince of Cairo, to be an exception, not an example. Baali could be Barons for sure and would as it helps them blend in to perform their evil acts. Banu I am unsure but since Banu are no longer whitelisted, it would be good to give them more jobs.

<details>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

## Changelog

:cl:
add: Added new mechanics or gameplay changes
add: Added more things
del: Removed old things
qol: made something easier to use
balance: rebalanced something
fix: fixed a few things
sound: added/modified/removed audio or sound effects
image: added/modified/removed some icons or images
map: added/modified/removed map content
spellcheck: fixed a few typos
code: changed some code
refactor: refactored some code
config: changed some config setting
admin: messed with admin stuff
/:cl:

### DIFF
--- a/code/modules/vtmb/jobs/anarchs/baron.dm
+++ b/code/modules/vtmb/jobs/anarchs/baron.dm
@@ -32,7 +32,7 @@
 	v_duty = "You lead the Anarchs in the City. A Camarilla faction that values freedom and individuality. While not a fan of the traditions, you still respect the Masquerade."
 	minimal_masquerade = 3
 	allowed_species = list("Vampire")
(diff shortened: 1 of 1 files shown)


Agateophobia (approved):
Sure, why not.

PR #944, opened 2025-03-13
Buff Disciplines

TheCarnalest:
## About The Pull Request
Increases physical Discipline duration from 6 seconds to 10 seconds, decreases Bloodheal cooldown from 6 seconds to 5 seconds, make the Galdjum artifact increase power duration again, make Auspex and Obfuscate fully toggled without needing to spend blood as in tabletop.

## Why It's Good For The Game
Makes physical Disciplines less awful to use, fixes a bug, and also brings Auspex and Obfuscate closer to tabletop. They still need rolling, but this is a STEP.

<details>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

## Changelog

:cl:
- Celerity, Fortitude, and Potence now last 10 seconds between spending vitae rather than 6.
- Bloodheal now has a 5 second cooldown rather than 6.
- Obfuscate and Auspex are now fully toggled and do not cost vitae, as in tabletop.
- Obfuscate's lower levels now require that you not be in anyone's sight to activate them. This does not apply to Obfuscate 4 and 5.
/:cl:

### DIFF
--- a/code/__DEFINES/wod13/discipline.dm
+++ b/code/__DEFINES/wod13/discipline.dm
@@ -1,8 +1,8 @@
 //normal duration defines
-///Duration of one "turn", which is 6 seconds according to us
-#define DURATION_TURN 6 SECONDS
-///Duration of one "scene", which is 5 minutes according to us
-#define DURATION_SCENE 5 MINUTES
+///Duration of one "turn", which is 5 seconds according to us
+#define TURNS * 5 SECONDS
+///Duration of one "scene", which is 3 minutes according to us
+#define SCENES * 3 MINUTES
 
 //targeting bitflags, NONE or 0 if targeting self exclusively
 ///Allows for self to also be selected in ranged targeting, SET TO 0 IF NOT TARGETED OR RANGED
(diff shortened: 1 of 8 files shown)


TheCarnalest:
I have reviewed my own code as the codeowner and found it to be flawless

TheCarnalest:
> Why not make Fortitude last longer then Celerity/Potence tabletop it should be entirely passive why not make it like a minute if not longer per cost spent.

This exclusively represents what happens when you activate Fortitude, namely massively reducing the amount of damage you take. Passive effects can be added, but these aren't supposed to be replacements. Also, someone keeping Fortitude 5 active for a minute would practically be unkillable because of armour stacking

YellowKingThe:
> > Why not make Fortitude last longer then Celerity/Potence tabletop it should be entirely passive why not make it like a minute if not longer per cost spent.
> 
> This exclusively represents what happens when you activate Fortitude, namely massively reducing the amount of damage you take. Passive effects can be added, but these aren't supposed to be replacements. Also, someone keeping Fortitude 5 active for a minute would practically be unkillable because of armour stacking

There is no activated form of Fortitude in tabletop until dot 6

